PR #17, opened 2018-07-08
Replaced all `py -3` to `python3`

chriscn:
Whether this is a local issue with my homebrew installation of Python but `py -3` doesn't work and so I have replaced it with python3 because that is more universal and works.


aliparlakci:
It is not universal for sure. Because when you install python 3 and you don't already have python 2, the prefix is "python", not python3. python3 works for you because python resembles Python 2

PR #75, opened 2019-12-23
Feature - User downvotes

KevinMidboe:
## Preface
New parameter for downloading downvoted posts for a user. I don't know if anyone else needs this parameter, but I found a use-case. 

Parses much the same as upvoted and uses the `reddit.redditor(args["user"]).downvoted` function to fetch downvoted posts. 

## Usage
Usage similar to *--upvoted* or *--saved*:
```bash
$: python ./script.py --downvoted
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ Downloads media from reddit posts. Made by [u/aliparlakci](https://reddit.com/u/
 
 ## What it can do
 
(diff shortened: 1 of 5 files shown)


aliparlakci:
I will have a deeper look into it. Thanks for the contribution!

Issue #122, opened 2020-06-07
Create --config option to explicitly specify config.json file to use.


r-pufky:
Currently bdfr pulls from {HOME}/Bulk downloader for reddit/config.json or a config in the local script directory. This makes usage difficult, especially if you have multiple configs that you use to pull via scripts and cronjobs.

I'd propose a `--config {FILE} option`, which explicitly points to the configuration to load. If it is not found, the program exits. It will also disable loading from local script directory and {HOME}/Bulk downloader for reddit/config.json. This will make running this in scripts and crontabs MUCH easier.

This wouldn't change how the program functions for basic users, and allow power users to easily use multiple configs without jumping through extra hoops.

I can probably put together a pull request if you are okay with it.

aliparlakci:
If you can open a pull request, I can merge it.

ymgenesis:
You could try using `--use-local-config` in the current directory, but make it a link/alias of a config.json file that exists somewhere else on your system. Not sure if that'd work, but Linux does this kind of thing all the time for its system config files and daemons.

PR #124, opened 2020-06-07
add --config (-c) option for specifying exact location of config.json to use.

r-pufky:
Added option per outline in #122 

* Added `--config` (`-c`) option, which accepts a file path to config.json.
* Added logic to verify that both --config and --use-local-config are not used at the same time.
* Updated documentation.

### DIFF
--- a/script.py
+++ b/script.py
@@ -256,14 +256,24 @@ def printLogo():
     )
 
 def main():
+    arguments = Arguments.parse()
+    GLOBAL.arguments = arguments
 
-    if not Path(GLOBAL.defaultConfigDirectory).is_dir():
-        os.makedirs(GLOBAL.defaultConfigDirectory)
-
-    if Path("config.json").exists():
-        GLOBAL.configDirectory = Path("config.json")
+    if arguments.config:
+        if arguments.use_local_config:
+            sys.exit()
+        if Path(arguments.config).exists():
+            GLOBAL.configDirectory = Path(arguments.config)
+        else:
+            VanillaPrint("custom config",arguments.config,"not found. Exiting.")
+            sys.exit()
     else:
-        GLOBAL.configDirectory = GLOBAL.defaultConfigDirectory  / "config.json"
+        if not Path(GLOBAL.defaultConfigDirectory).is_dir():
+            os.makedirs(GLOBAL.defaultConfigDirectory)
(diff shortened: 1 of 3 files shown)


aliparlakci:
Can you remove the 3d70fb9 commit?

r-pufky:
Sure I'll do it later tonight and update.

r-pufky:
Should be all set. I rebase'd and removed the whitespace cleanup.

script.py
```diff
@@ -256,14 +256,20 @@ def printLogo():
...
+    arguments = Arguments.parse()
+    GLOBAL.arguments = arguments
 
-    if not Path(GLOBAL.defaultConfigDirectory).is_dir():
-        os.makedirs(GLOBAL.defaultConfigDirectory)
-
-    if Path("config.json").exists():
-        GLOBAL.configDirectory = Path("config.json")
+    if arguments.config:
+        if arguments.use_local_config:
```
aliparlakci, line 263: Why does it exit if `--use-local-config` is given?

script.py
```diff
@@ -256,14 +256,20 @@ def printLogo():
...
+            sys.exit()
+        GLOBAL.configDirectory = Path(arguments.config)
```
aliparlakci, line 265: There should be a validation that the given file actually exists.

src/arguments.py
```diff
@@ -101,6 +101,7 @@ def parse(arguments=[]):
                             metavar="TIME_LIMIT",
                             type=str)
 
+
```
aliparlakci, line 104: Remove this extra line here.

script.py
r-pufky: These options inherently conflict with each other, with `--use-local-config` creating a new config if it doesn't exist in current directory, as well as loading it from the current directory.

Specifying `--config` implicitly means that you've already created the config or are loading it from a known location.

Happy to remove the check here. I have no strong opinions either way :)

src/arguments.py
r-pufky: done.

script.py
r-pufky: Added validation with log message.

script.py
r-pufky: Hrmm, it's been two months since your comment with no responses. I'm going to resolve this.

PR #176, opened 2020-12-28
Added new --download-delay flag to allow throttling downloads

dwbfox:
Added new --download-delay flag to allow throttling the download across all supported platforms. This is useful for certain platforms like Reddit that will begin throwing 403 Forbidden responses after a sufficient amount of time has elapsed downloading images.

- added a new cli argument called `--download-delay` to specify the number of int seconds to delay each download
- added a check in scripts.py to get the passed argument value to delay the download by
- this delay would only come into effect on successful downloads (where the download counter would have been incremented)

### DIFF
--- a/script.py
+++ b/script.py
@@ -36,6 +36,8 @@ from src.programMode import ProgramMode
 from src.reddit import Reddit
 from src.store import Store
 
+from time import sleep
+
 __author__ = "Ali Parlakci"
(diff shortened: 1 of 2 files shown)


dwbfox:
uhh...anyone home to review this request? 👀

aliparlakci (approved):
LGTM

PR #177, opened 2021-01-07
Reddit native album downloader, minor fix for imgur json error, access denied in redgifs

EkriirkE:
This is literally my first time playing in Python, so code is not going to be clean or could be done in other easier/efficient ways.
The album downloader was modified from the Imgur script - it can be smaller but I tried to stay with the original style.

Added a small tweak for imgur that was randomly throwing json parsing errors due to an extra comma that should be a close curly so I let it try the original method first then try mine if it fails


aliparlakci (approved):
LGTM

PR #184, opened 2021-02-06
Pep8 format

### LINKED ISSUE #183: Format code to comply with PEP8 standard

Serene-Arc:
Closes #183 

Sweeping changes to the code but little of actual substance; mostly this was just formatting the code to make it easier to refactor and merge in changes in the future.

The changes are pretty simple:
	- Formatting each file according to the PEP8 standard with autopep8
		- The sole exception is a line length of 120 instead of 80 because I think 80 is stupid
	- Sorting imports and removing any unused imports
	- Condensing some multiline code

For the most part, each change is confined to a single file. I'll be doing a broader refactor soon but in order to keep the git diffs smaller and to keep more atomic commits, all of the autoformatting has been done here.

### DIFF
--- a/src/errors.py
+++ b/src/errors.py
@@ -1,9 +1,11 @@
 import sys
 
+
 def full_exc_info(exc_info):
 
     def current_stack(skip=0):
-        try: 1/0
+        try:
+            1 / 0
         except ZeroDivisionError:
             f = sys.exc_info()[2].tb_frame
         for i in range(skip + 2):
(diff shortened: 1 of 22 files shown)


Serene-Arc:
Also this would be best squashed because there's a lot of small commits for each individual file

src/errors.py
```diff
@@ -33,80 +35,102 @@ def __init__(self, tb_frame, tb_lineno, tb_next):
...
+
 class InsufficientPermission(Exception):
     pass
 
+
 class InvalidSortingType(Exception):
     pass
 
-class FileNotFoundError(Exception):
-    pass
```
aliparlakci, line 85: I assume you removed this error as it is not referenced anywhere?

aliparlakci (changes requested):
Great!

Do you think we should make refactoring on a different branch? I created `develop` branch from master and we can merge this into there instead of master.

src/errors.py
Serene-Arc: FileNotFoundError is an error that is built into Python, so by defining it there, it was overwriting something that already exists. For an exception, it *probably* doesn't matter too much doing that, but in this case it doesn't seem to be necessary. Still, one of the things I have planned for the refactoring is to remove all of the shadowed functions because they can be confusing

PR #188, opened 2021-02-06
Add typing to code

### LINKED ISSUE #185: Implement typing
Add typing annotations to the code

Serene-Arc:
Closes #185 

Adds some typing to most functions. It may not be all correct, but it will be verified during the refactor anyway, I just wanted to do this step as a PR before beginning.

### DIFF
--- a/src/downloaders/Direct.py
+++ b/src/downloaders/Direct.py
@@ -1,11 +1,12 @@
 import os
+import pathlib
 
 from src.downloaders.downloaderUtils import getExtension, getFile
 from src.utils import GLOBAL
 
 
 class Direct:
-    def __init__(self, directory, post):
+    def __init__(self, directory: pathlib.Path, post: dict):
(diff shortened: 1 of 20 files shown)


vladdoster:
Could you point out which parts you arent sure about the correctness since it is a large diff.

src/downloaders/Direct.py
```diff
@@ -1,11 +1,12 @@
 import os
+import pathlib
 
 from src.downloaders.downloaderUtils import getExtension, getFile
 from src.utils import GLOBAL
 
 
 class Direct:
-    def __init__(self, directory, post):
+    def __init__(self, directory: pathlib.Path, post: dict):
```
vladdoster, line 9: wouldn't the following be more appropriate?

```
typing.Union[str, pathlib.Path]
```

Serene-Arc:
Sorry, but this pull request has been rendered obsolete by a full-rewrite that I'm in the process of completing now.

PR #201, opened 2021-03-21
Renamed gallery.py to albumGallery.py to allow script to run on Windows

bad-jacob:
Because of the case-insensitive nature of Windows, the script doesn't currently run there.

I've removed the reference to lowercase gallery because it isn't being used anyway.


Serene-Arc:
Issue was resolved under PR #203 which has been merged. Thanks though :)

PR #204, opened 2021-03-22
Bug fixes

Serene-Arc:
Just a couple quick bug fixes:

- Fix some typos in the README
- Copy the default config to the config directory if none found
- Filter invalid characters from Windows paths

### DIFF
--- a/bulkredditdownloader/file_name_formatter.py
+++ b/bulkredditdownloader/file_name_formatter.py
@@ -2,6 +2,7 @@
 # coding=utf-8
 
 import logging
+import platform
 import re
 from pathlib import Path
 from typing import Optional
@@ -41,6 +42,10 @@ class FileNameFormatter:
                 logger.log(9, f'Found key string {key} in name')
 
         result = result.replace('/', '')
+
+        if platform.system() == 'Windows':
+            result = FileNameFormatter._format_for_windows(result)
+
         return result
 
     def format_path(self, resource: Resource, destination_directory: Path, index: Optional[int] = None) -> Path:
@@ -51,6 +56,7 @@ class FileNameFormatter:
         ending = index + resource.extension
         file_name = str(self._format_name(resource.source_submission, self.file_format_string))
         file_name = self._limit_file_name_length(file_name, ending)
+
         try:
             file_path = Path(subfolder, file_name)
         except TypeError:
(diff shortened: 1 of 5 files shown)


aliparlakci:
I will review it and merge it. Please don't merge it without my review.

bulkredditdownloader/file_name_formatter.py
```diff
@@ -41,6 +42,10 @@ def _format_name(submission: praw.models.Submission, format_string: str) -> str:
                 logger.log(9, f'Found key string {key} in name')
 
         result = result.replace('/', '')
+
+        if platform.system() == 'Windows':
+            result = FileNameFormatter._format_for_windows(result)
+
```
aliparlakci, line 48: We should not format file names platform conditionally. Because, bdfr is commonly used in headless servers which run on Linux. However, when transfering files to a personal computer running Windows, they will cause problems.

On the other hand, right now, it is throwing file name issues because of Unicode characters (emojis, mainly).

So, I suggest replacing _**all**_ non-ASCII characters in addition to the ones that are problematic to Windows.

aliparlakci (changes requested):

bulkredditdownloader/file_name_formatter.py
Serene-Arc: I disagree, there's no reason to discard information because of the poor design of Windows and the archaic filesystem that it uses. There are many Linux tools to clean file names to be compatible with Windows, and many implementations of SAMBA and otherfile sharing protocols from Linux to Windows, include this functionality. When emojis and regular punctuation are used in the titles of posts, the BDFR should be able to accurately save this information unless doing so will crash the program.

aliparlakci: Emojis crash program on Windows.

aliparlakci (changes requested):

Serene-Arc:
The changes requested can be discussed and merged in another request, I'm going to merge this one so that the beta is functional on Windows until a decision is made.

aliparlakci:
I wrote a comment here but I now realise that it was not send. 

> The changes requested can be discussed and merged in another request, I'm going to merge this one so that the beta is functional on Windows until a decision is made.

I agree

PR #227, opened 2021-03-27
File formatter bug fix

aliparlakci:
string.strip() does not strictly match the whole word. So, this causes some of the unrelated parts of the filename to be stripped. Replaced it with regex.

Also, string.format() can be used to construct a filename using template instead of doing regex. A good advantage of using curly brackets in schemes :)

### DIFF
--- a/bulkredditdownloader/file_name_formatter.py
+++ b/bulkredditdownloader/file_name_formatter.py
@@ -27,15 +27,20 @@ class FileNameFormatter:
     @staticmethod
     def _format_name(submission: praw.models.Submission, format_string: str) -> str:
         submission_attributes = {
-            'TITLE': submission.title,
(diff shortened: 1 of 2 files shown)


Serene-Arc (changes requested):
Denied

Serene-Arc:
I also fail to see the reasoning behind using regex in 61b3102, what is the case where there is a problem with it? Also, for future reference, if you do find a case where the code misbehaves, please add a test that would catch that behaviour in the future.

aliparlakci:
I will revert 8e82252 and add a test for 61b3102

Serene-Arc:
I added a commit on the main v2 branch for a test that shows the non-existent key error, so maybe rebase off of that too

PR #230, opened 2021-03-27
Send HTTP200 when oauth2 is successful

aliparlakci:
Browser waits for HTTP200 response after OAuth is successful. Also, a feedback added to the response.

### DIFF
--- a/bulkredditdownloader/oauth2.py
+++ b/bulkredditdownloader/oauth2.py
@@ -62,6 +62,7 @@ class OAuth2Authenticator:
             self.send_message(client)
             raise RedditAuthenticationError(f'Error in OAuth2: {params["error"]}')
 
+        self.send_message(client, "<script>alert('You can go back to terminal window now.')</script>")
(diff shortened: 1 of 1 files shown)


Serene-Arc:
I don't think this is necessary, the BDFR already sends a HTTP 200 response, and the feedback from the successful receipt of a token is that the BDFR will continue and begin execution.

aliparlakci:
> the BDFR already sends a HTTP 200 response

it does not. self.send_message() should be called before return statement to return 200. you can check the network tab on DevTools to see failing the request

Also, I don't see any harm on sending an alert along with the successful response. Because on an ordinary OAuth flow, newly opened tab or window would close and you are redirected to the main site. In our case, it just hangs on an empty page. More specifically on an error page as if something went wrong because we are not sending HTTP20x.

Serene-Arc (approved):
Very well, if you think that it's a benefit for the BDFR

PR #233, opened 2021-03-28
Add warning for non-unique file name schemes

### LINKED ISSUE #216: Display a warning message before starting if filenames are not guaranteed to be unique.
If {POSTID} does not exists on the filename scheme, for two separate posts with same titles, the target filenames might be the same. Then BDfR will only download one of them and won't download others as can be seen below. So, a message should be displayed warning the user about such possible behavior before starting.

Also, this occasion should be stated on README clearly

image

Serene-Arc:
Closes #216 

Adds a logger warning when using a file scheme that cannot be guaranteed to be unique

### DIFF
--- a/bulkredditdownloader/file_name_formatter.py
+++ b/bulkredditdownloader/file_name_formatter.py
@@ -89,7 +89,14 @@ class FileNameFormatter:
     def validate_string(test_string: str) -> bool:
         if not test_string:
             return False
-        return any([f'{{{key}}}' in test_string.lower() for key in FileNameFormatter.key_terms])
+        result = any([f'{{{key}}}' in test_string.lower() for key in FileNameFormatter.key_terms])
+        if result:
+            if 'POSTID' not in test_string:
+                logger.warning(
+                    f'Post ID not included in this file scheme, so file names are not guaranteed to be unique')
+            return True
+        else:
+            return False
 
     @staticmethod
     def _format_for_windows(input_string: str) -> str:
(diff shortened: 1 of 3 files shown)


bulkredditdownloader/file_name_formatter.py
```diff
@@ -89,7 +89,14 @@ def format_resource_paths(self, resources: list[Resource],
     def validate_string(test_string: str) -> bool:
         if not test_string:
             return False
-        return any([f'{{{key}}}' in test_string.lower() for key in FileNameFormatter.key_terms])
+        result = any([f'{{{key}}}' in test_string.lower() for key in FileNameFormatter.key_terms])
+        if result:
+            if 'POSTID' not in test_string:
+                logger.warning(
+                    f'Post ID not included in this file scheme, so file names are not guaranteed to be unique')
```
aliparlakci, line 96: "File names are not guaranteed to be unique as Post IDs are not included in the file scheme. Some files might not be downloaded due to name conflicts."

aliparlakci (changes requested):
Program behavior should be explained in [README/Folder and File Name Schemes](https://github.com/aliparlakci/bulk-downloader-for-reddit/tree/v2#folder-and-file-name-schemes) also.

Serene-Arc:
The warning has been added to the README

aliparlakci (approved):

bulkredditdownloader/file_name_formatter.py
aliparlakci: @Serene-Arc

aliparlakci (changes requested):

Issue #238, opened 2021-03-30
Optimize tests so that they can be run on a CI provider for every commit


aliparlakci:
Some of the test cases heavily depend on requests to outer sources and consumes too much bandwidth which creates a bottleneck in the test process. Especially reddit takes too long to respond to the requests. They should optimized in order to add CI steps to repository.

This is a general issue which test optimization pull requests can give reference to.

Serene-Arc:
For some test cases that might be possible to do, but it's also the reason that I included the marks system. You can exclude all tests that contact external resources, or tests that contact Reddit, or tests that require an authentication token, or any permutation of them all.

Serene-Arc:
To clarify, it will *never* be possible to run every test with a CI provider. Some tests run for a long time regardless, and some require an authentication token that can't really be provided in this kind of environment. We can run most tests however.

PR #239, opened 2021-03-30
test_download_factory.py: hardcode submission links

aliparlakci:
As post id's are only required to get content url, they can be replaced by actual URLs. This way, test_download_factory.py is 98% faster. Related to #238.

image


Serene-Arc (changes requested):
If you're no longer requesting resources directly from Reddit, and no longer making an external network request, then the two marks at the top of the test have to be remove. They're what classify this test as requiring those things.

Serene-Arc (approved):

Issue #218, opened 2021-03-27
V2: Unable to run setup.py


koeningyou:
Hi,

I'm unable to run BDFRv2 setup. I get the error below when running "python3 setup.py";

```
ERROR:root:Error parsing
Traceback (most recent call last):
  File "/home/XXXX/Downloads/bulk-downloader-for-reddit-2/.eggs/pbr-5.5.1-py3.8.egg/pbr/core.py", line 96, in pbr
    attrs = util.cfg_to_args(path, dist.script_args)
  File "/home/XXXX/Downloads/bulk-downloader-for-reddit-2/.eggs/pbr-5.5.1-py3.8.egg/pbr/util.py", line 271, in cfg_to_args
    pbr.hooks.setup_hook(config)
  File "/home/XXXX/Downloads/bulk-downloader-for-reddit-2/.eggs/pbr-5.5.1-py3.8.egg/pbr/hooks/__init__.py", line 25, in setup_hook
    metadata_config.run()
  File "/home/XXXX/Downloads/bulk-downloader-for-reddit-2/.eggs/pbr-5.5.1-py3.8.egg/pbr/hooks/base.py", line 27, in run
    self.hook()
  File "/home/XXXX/Downloads/bulk-downloader-for-reddit-2/.eggs/pbr-5.5.1-py3.8.egg/pbr/hooks/metadata.py", line 25, in hook
    self.config['version'] = packaging.get_version(
  File "/home/XXXX/Downloads/bulk-downloader-for-reddit-2/.eggs/pbr-5.5.1-py3.8.egg/pbr/packaging.py", line 874, in get_version
    raise Exception("Versioning for this project requires either an sdist"
Exception: Versioning for this project requires either an sdist tarball, or access to an upstream git repository. It's also possible that there is a mismatch between the package name in setup.cfg and the argument given to pbr.version.VersionInfo. Project name Bulk Downloader for Reddit was given, but was not able to be found.
error in setup command: Error parsing /home/farooq/Downloads/bulk-downloader-for-reddit-2/setup.cfg: Exception: Versioning for this project requires either an sdist tarball, or access to an upstream git repository. It's also possible that there is a mismatch between the package name in setup.cfg and the argument given to pbr.version.VersionInfo. Project name Bulk Downloader for Reddit was given, but was not able to be found.
```

Serene-Arc:
What happens if you try and install through `python3 -m pip install .` in the root folder?

[...]

PR #241, opened 2021-03-31
instructions for v2 build

### LINKED ISSUE #240: No documenation for installing v2 locally
I am not sure how to create the build after doing a clone and a pip requirements.

```
~/bulk-downloader-for-reddit$ python3 setup.py 
usage: setup.py [global_opts] cmd1 [cmd1_opts] [cmd2 [cmd2_opts] ...]
   or: setup.py --help [cmd1 cmd2 ...]
   or: setup.py --help-commands
   or: setup.py cmd --help

error: no commands supplied
```

cgsmith:
Fixes #240 

Related to #218


Serene-Arc:
I believe that this PR is premature. This part of the README will be added when the beta is over and will link to the pypi package that is downloadable with pip. That is preferable for the bulk of people, and these instructions are not hard to find in any case as this is the standard way to install a package from source.

cgsmith:
@Serene-Arc if the instructions were not necessary this issue wouldn't pop up again. A lot of developers may want to contribute but aren't familiar with Python's build process (like me). Having some documentation encourages future growth and participation from a larger community. 

I can put this in a contribution section.

Serene-Arc:
I think the contributing document would be a better place for it. I don't want regular users getting confused and installing this way because it's not optimal for a couple of reasons. But if developers need it then the contributing docs would be the place. 

Also the best command would be ‘python3 -m pip install -e .’ This installs the package while remaining linked to the source so any changes to the code can be tested without reinstalling.

aliparlakci:
This PR can be merged if the necessary steps to setup the environment for development are provided in the contributions doc.

mchubby:
Not a build error per se, but it seems Python 3.9  is the minimal requirement for running the script.

I get this error from a 3.8 venv:
TypeError: 'type' object is not subscriptable

Serene-Arc:
Yes I believe there were a couple typing hints that are only available in 3.9 onwards

PR #244, opened 2021-04-02
Add logic to handle mobile youtube (m.youtube) links.

BlipRanger:
Quick edit to support mobile youtube links. Discovered while tested and confirmed to work against m.youtube links.

### DIFF
--- a/bulkredditdownloader/site_downloaders/download_factory.py
+++ b/bulkredditdownloader/site_downloaders/download_factory.py
@@ -40,9 +40,7 @@ class DownloadFactory:
             return SelfPost
         elif re.match(url_beginning + r'v\.redd\.it', url):
             return VReddit
-        elif re.match(url_beginning + r'youtu\.?be', url):
-            return Youtube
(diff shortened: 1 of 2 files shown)


aliparlakci:
Necessary tests should also be added

aliparlakci (changes requested):
Necessary tests should also be added

Serene-Arc:
This needs tests, also please modify the existing YouTube regex to avoid adding another entry that is superfluous. Since the factory is likely to grow, keeping it as small as is feasible would be good. `r'(m\.)?youtu\.?be` would do the same thing in the same entry

Serene-Arc (changes requested):
Also please modify the existing YouTube regex to avoid adding another entry that is superfluous. Since the factory is likely to grow, keeping it as small as is feasible would be good. `r'(m\.)?youtu\.?be` would do the same thing in the same entry

BlipRanger:
Alright, made that small change and added a test. Was there another one needed or does that cover it?

Serene-Arc (approved):
Looks fine to me, merging

PR #250, opened 2021-04-03
Control direct links before domain names

aliparlakci:
If a URL ends with an extension, consider it a direct link instead of searching for domains. Without this, it fails to recognize direct content in links such as https://imgur.com/3ls94yv.jpeg


Serene-Arc (changes requested):
I don't think is this a good idea, the default being that any url with an explicit extension being last was done for a reason. Some sites do not actually return the file in those cases, but another webpage that has to be parsed to get the actual resource. Hence, that regex being last to catch those links as a last resort if there aren't any specific downloaders available to catch it before then. The gifv situation with Imgur is a prime example of that

Serene-Arc:
Also why did you merge from v2 into this branch?

aliparlakci:
> Also why did you merge from v2 into this branch?

v2 was ahead of this branch and it was causing merge conflicts. I merged v2 locally and fixed the conflicts.

aliparlakci:
> I don't think is this a good idea, the default being that any url with an explicit extension being last was done for a reason. Some sites do not actually return the file in those cases, but another webpage that has to be parsed to get the actual resource. Hence, that regex being last to catch those links as a last resort if there aren't any specific downloaders available to catch it before then. The gifv situation with Imgur is a prime example of that

Then, there are two ways to solve the above problem. We either check if it is a direct link right away with the exceptions such as ".gifv" or we add also the extension regexes for every domain. I was originally going to implement the latter but I couldn't think any case where url ends with an extension but it is not a direct link, other than ".gifv", among the known domains. If the domain is not known, then we have no choice but to assume it is a direct link.

One solution is simple but includes assumptions (which I think are reasonable), other one might complicate the download_factory but it makes no assumptions. You can pick which way we should go. But we must pick one because we cannot download the above link right now.

Serene-Arc:
Why would we need to add extension regex to every domain name? There is a third alternative, in which the Imgur link can handle and pass this to the Direct class if needed

Serene-Arc:
Okay on further thought I think that with tests, these assumptions should be fine. Merging

aliparlakci:
> Why would we need to add extension regex to every domain name? There is a third alternative, in which the Imgur link can handle and pass this to the Direct class if needed.

Calling the appropriate downloader class is the responsibility of download_factory. Downloader calling another downloader would easily get things more complicated eventually making download_factory useless.

Serene-Arc:
Well funny you should say that but I actually just removed a bunch of legacy code that did just that. After merging this, I removed a bunch of logic that just returned the given URL if it was a direct link. So *now* it doesn't do it but before, that was the standard way that the downloaders handled it, way back from v1

aliparlakci:
If v1 was good, there wouldn't be v2 :D v2 is supposed to be better than v1.

I was aware there were things which violate what I said. But there is no point in continuing a mistake. Otherwise, it will make a snowball effect.

All these efforts is for not having a need to reimplement program again in the future. :)

Serene-Arc:
Well I did remove them so moving forwards, any exception to the rule will have to be coded in manually.

PR #256, opened 2021-04-06
Add //imgur.com/*.gifv

### LINKED ISSUE #255: Doesn't download gifv from imgur if url is not i.imgur
Idk if this is the right spot to put user issues, I just noticed that it would download gifv's fine if the url was https://i.imgur.com/xxxxxxx.gifv but not if it was https://imgur.com/xxxxxxx.gifv, which some posts are.

thanks.

This is for v2 btw

aliparlakci:
Closes #255


Serene-Arc:
I need to code a fix for this because if you look at the actual Imgur module, it fails to parse this kind of link at all. I'll find a fix for that and push to this branch before merging.

PR #264, opened 2021-04-13
Update extension regex to match URI fragments

### LINKED ISSUE #266: Downloads error if URI fragment present
On posts where the a user incorrectly adds a fragment to the URI of a direct file, the download fails.

Example:
```
python3 -m bdfr download -l mpsrds /tmp -v
[2021-04-12 22:44:36,195 - bdfr.downloader - DEBUG] - Loading configuration from /home/froppy/.config/bdfr/default_config.cfg
[2021-04-12 22:44:36,195 - bdfr.downloader - DEBUG] - Setting maximum download wait time to 120 seconds
[2021-04-12 22:44:36,196 - bdfr.downloader - DEBUG] - Using unauthenticated Reddit instance
[2021-04-12 22:44:36,196 - bdfr.downloader - DEBUG] - Attempting to download submission mpsrds
[2021-04-12 22:44:36,508 - bdfr.downloader - DEBUG] - Using Direct with url https://i.redd.it/gtz9yxa4lus61.jpg#test
Traceback (most recent call last):
  File "/usr/lib/python3.9/runpy.py", line 197, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib/python3.9/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/home/froppy/src/bulk-downloader-for-reddit/bdfr/__main__.py", line 117, in <module>
    cli()
  File "/home/froppy/.local/lib/python3.9/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/home/froppy/.local/lib/python3.9/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/home/froppy/.local/lib/python3.9/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/froppy/.local/lib/python3.9/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/froppy/.local/lib/python3.9/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/home/froppy/.local/lib/python3.9/site-packages/click/decorators.py", line 21, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "/home/froppy/src/bulk-downloader-for-reddit/bdfr/__main__.py", line 64, in cli_download
    reddit_downloader.download()
  File "/home/froppy/src/bulk-downloader-for-reddit/bdfr/downloader.py", line 361, in download
    self._download_submission(submission)
  File "/home/froppy/src/bulk-downloader-for-reddit/bdfr/downloader.py", line 383, in _download_submission
    for destination, res in self.file_name_formatter.format_resource_paths(content, self.download_directory):
  File "/home/froppy/src/bulk-downloader-for-reddit/bdfr/file_name_formatter.py", line 135, in format_resource_paths
    out.append((self.format_path(resources[0], destination_directory, None), resources[0]))
  File "/home/froppy/src/bulk-downloader-for-reddit/bdfr/file_name_formatter.py", line 105, in format_path
    raise BulkDownloaderException(f'Resource from {resource.url} has no extension')
bdfr.exceptions.BulkDownloaderException: Resource from https://i.redd.it/gtz9yxa4lus61.jpg#test has no extension
```
This also blocks further downloads from happening if downloaded from a subreddit for example since bdfr exits.

Boo1098:
Fixed issues when end of link has an html id in it.

I found that (rarely) there were images that people had added an html tag to the end of so the link would look like `https://i.redd.it/imageid.jpg#htmltag` and would cause an error as it would not be able to recognize an extension for the file. This fixes that issue.

Fixes #266

### DIFF
--- a/bdfr/resource.py
+++ b/bdfr/resource.py
@@ -64,7 +64,7 @@ class Resource:
         self.hash = hashlib.md5(self.content)
 
     def _determine_extension(self) -> str:
-        extension_pattern = re.compile(r'.*(\..{3,5})(?:\?.*)?$')
+        extension_pattern = re.compile(r'.*(\..{3,5})(?:(?:\?|#).*)?$')
         match = re.search(extension_pattern, self.url)
         if match:
             return match.group(1)
--- a/bdfr/resource.py
+++ b/bdfr/resource.py
@@ -64,7 +64,7 @@ class Resource:
         self.hash = hashlib.md5(self.content)
 
     def _determine_extension(self) -> str:
-        extension_pattern = re.compile(r'.*(\..{3,5})(?:(?:\?|#).*)?$')
+        extension_pattern = re.compile(r'.*(\..{3,5})(?:\?.*)?(?:#.*)?$')
         match = re.search(extension_pattern, self.url)
         if match:
             return match.group(1)
--- a/bdfr/tests/test_downloader.py
+++ b/bdfr/tests/test_downloader.py
@@ -268,7 +268,6 @@ def test_get_user_authenticated_lists(
 @pytest.mark.reddit
 @pytest.mark.parametrize(('test_submission_id', 'expected_files_len'), (
     ('ljyy27', 4),
-    ('mpsrds', 1),
 ))
 def test_download_submission(
         test_submission_id: str,
--- a/bdfr/tests/test_downloader.py
+++ b/bdfr/tests/test_downloader.py
@@ -268,6 +268,7 @@ def test_get_user_authenticated_lists(
 @pytest.mark.reddit
 @pytest.mark.parametrize(('test_submission_id', 'expected_files_len'), (
     ('ljyy27', 4),
+    ('mpsrds', 1),
 ))
 def test_download_submission(
         test_submission_id: str,
--- a/bdfr/tests/test_resource.py
+++ b/bdfr/tests/test_resource.py
@@ -15,6 +15,8 @@ from bdfr.resource import Resource
     ('https://www.resource.com/test/example.jpg', '.jpg'),
     ('hard.png.mp4', '.mp4'),
     ('https://preview.redd.it/7zkmr1wqqih61.png?width=237&format=png&auto=webp&s=19de214e634cbcad99', '.png'),
+    ('test.jpg#test','.jpg'),
+    ('test.jpg?width=247#test','.jpg'),
 ))
 def test_resource_get_extension(test_url: str, expected: str):
     test_resource = Resource(MagicMock(), test_url)


Serene-Arc (changes requested):
Please add test cases to the test suite to prove that it does what it's supposed to and doesn't make other tests break. 

Also, it's almost fine, but the way it is at the moment means that it will match for one set of URL parameters OR one URI fragment, but not both. To make it that, change it to the following (the last question mark changed):

`r'.*(\..{3,5})(?:(?:\?|#).*)*$'`

Also, I'm curious what types of links that we allow in the downloaders are being caught out by this. Could you please provide a couple so I can examine them and write some test cases for the relevant site downloader module?

Serene-Arc:
Actually `{,2}` would be a better fit instead of a star, let's not be overly greedy. That would make it:

```regex
.*(\..{3,5})(?:(?:\?|#).*){,2}$
```

Serene-Arc:
On second thought, I think it would be better to have a properly descriptive regex of a URL, since the URI fragment always comes after the URL parameters, we can catch it in a separate group. Try something along the lines of:

```regex
.*(\..{3,5})(?:\?.*)?(?:#.*)?$
```

Boo1098:
I set up a test account with some examples. I also found it interesting that it couldn't download the direct link to the jpg from nasa (which according to the docs it should be able to grab).

To test them all use:
```python
python3 -m bdfr download -s u_bdfr_throaway /tmp -v
```

The specific one with the case this fixes is `mpsrds`

```python
python3 -m bdfr download -l mpsrds /tmp -v
```

Technically no link should ever look like this since the html tag does nothing, but it could. And since it causes an unhandled exception it stops the current download process.

Also sorry for not reading the contributing doc and not creating an issue before posting.

I'll make the regex more robust and add tests.

Serene-Arc:
It's fine, glad to have you contributing :) That test that you gave isn't what I meant, if you look in `tests/test_resource.py` then you'll find a test named `test_resource_get_extension`. This is the test that checks whether URLs are being passed by the regex correctly. If you could put your test case here and delete the one you added in `test_downloader.py` that would be great. The downloader test isn't really necessary since this is the part in question and adding another online test isn't the best thing in this case.

Serene-Arc:
Also, just a heads up, you don't need to provide a throwaway on Reddit :) Providing me the logs or even the submission ID is enough for me to be able to replicate, test, and create a solution if needed.

Boo1098:
> Also, just a heads up, you don't need to provide a throwaway on Reddit :) Providing me the logs or even the submission ID is enough for me to be able to replicate, test, and create a solution if needed.

Lol I couldn't find the one that caused the issue so I just created my own.

Serene-Arc:
Fair lol, looks good to me

PR #280, opened 2021-04-19
V2

aliparlakci:


Serene-Arc (changes requested):
This is a pointless change, the BDFR is only compatible with Python 3.9 @vladdoster . This commit should be reverted

vladdoster:
> This is a pointless change, the BDFR is only compatible with Python 3.9 @vladdoster . This commit should be reverted

Why is 3.9 only supported? That is somewhat of a poor design choice given EOL support for the lower versions.

| version | supported through |
|---------|-------------------|
| 3.6     | 2021              |
| 3.7     | 2023              |
| 3.8     | 2024              |

Was there a discussion somewhere I could review the rationale for only supporting 3.9? 

BDFR optimally should support versions as low as the default Python 3 version bundled with common Linux distributions.

Serene-Arc:
This has been the case since the beginning of v2's development because we are using features that were only introduced in 3.9. You haven't been involved in the development of v2. It's not pinned to a single version, it's pinned to a minimum version, which is 3.9. Python 3.9 exists in every common distro repository and is about to become the default on the Ubuntu family within the next few weeks.

Please don't add commits to branches that have not been discussed with the people involved. This wasn't your branch and wasn't supposed to be open to changes. This was simply an easier method to reorganise the project history before switching over to v2 completely. It is not the place for additional commits, especially ones like this.

aliparlakci:
I have removed versions below Python 3.9 so the pipeline passes. This particular matter can be discussed on a separate issue.

PR #287, opened 2021-04-20
setup.py: forbid installing below 3.9

aliparlakci:
Currently, pip can install the package to environments with Python below 3.9. This PR forbids such action.

However, there might be other ways to achieve this behavior so it is open to discussion.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,4 +3,4 @@
 
 from setuptools import setup
 
-setup(setup_requires=['pbr', 'appdirs'], pbr=True, data_files=[('config', ['bdfr/default_config.cfg'])])
+setup(setup_requires=['pbr', 'appdirs'], pbr=True, data_files=[('config', ['bdfr/default_config.cfg'])], python_requires='>=3.9.0')
--- a/setup.cfg
+++ b/setup.cfg
@@ -4,7 +4,7 @@ description_file = README.md
 description_content_type = text/markdown
 home_page = https://github.com/aliparlakci/bulk-downloader-for-reddit
 keywords = reddit, download, archive
-version = 2.0.3
+version = 2.0.0
 author = Ali Parlakci
 author_email = parlakciali@gmail.com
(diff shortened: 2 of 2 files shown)


setup.py
```diff
@@ -1,6 +1,11 @@
...
 
+import sys
+
 from setuptools import setup
 
-setup(setup_requires=['pbr', 'appdirs'], pbr=True, data_files=[('config', ['bdfr/default_config.cfg'])])
+if sys.version_info >= (3,9):
+    setup(setup_requires=['pbr', 'appdirs'], pbr=True, data_files=[('config', ['bdfr/default_config.cfg'])])
+else:
+    raise RuntimeError("This package requres Python 3.9+")
```
vladdoster, line 11: requires

Serene-Arc (changes requested):
This isn't necessary, we can add a single line to `setup.cfg` as per [here](https://opendev.org/openstack/pbr/src/branch/master/setup.cfg). Just adding `python-requires = >=3.9` would work fine

setup.py
Serene-Arc: Did your message get cut off? This makes no sense

aliparlakci:
> This isn't necessary, we can add a single line to `setup.cfg` as per [here](https://opendev.org/openstack/pbr/src/branch/master/setup.cfg). Just adding `python-requires = >=3.9` would work fine

The line is already there and it seems to be not working. Check out the pypi page, Python requirement is listed there as >= 3.9 as well but I was able to insall it on Python 3.8

setup.py
aliparlakci: There is a typo in the error message @Serene-Arc

Serene-Arc:
I vote for the `setup()` parameter then, rather than this whole try-except statement

aliparlakci:
I will try having it on setup.py with TestPyPI see if it rejects installing.

PR #289, opened 2021-04-20
tests: move outside of the package

aliparlakci:
Move tests outside to the root directory.


Serene-Arc (approved):
Looks good to me, there's no difference to it changing places I think.

PR #290, opened 2021-04-20
github actions: test on windows and macos

aliparlakci:


Serene-Arc:
Ah...okay, this is somewhat unexpected. Were you running into these problems yourself when running the tests on Windows @aliparlakci? 

The reason for a lot of these tests failing is that `pathlib` detects the OS and changes the path's representation accordingly. I wrote all the expected results in Linux format, because that's what I'm running and also because it's the most common (Linux and MacOS). Windows uses the other system with backslashes instead of forward-slashes. That's why a lot of the unit tests are failing. No idea for the integration tests.

Is this a quirk of GitHub or did all these tests fail on your system too and I just didn't know?

aliparlakci:
@Serene-Arc #292

Issue #321, opened 2021-05-01
[FEATURE] Customize the file scheme {DATE} parameter


DennisPing:
- [x] I am requesting a feature.
- [x] I am running the latest version of BDfR
- [x] I have read the [Opening an issue](README.md#configuration)
  
## Description
In the `development` branch, the {DATE} key returns a full ISO 8601 datetime. For example: `2021-04-30T09:20:45`.
This full ISO format is the defacto standard which is great for databases, scientists, archivists. Additionally, I would like the option to shorten or customize this datetime. For full customizability, allow the user to call `%Y, %m, %d, %H, %M, %S`.
```
Full ISO datetime: 2021-04-30T09:20:45
Short ISO datetime: 2021-04-30 (this is just a date object now; not datetime object)
```

## Configuration file
It makes sense to set the date scheme in the config.cfg file.

## End Result
The CLI command below would operate the same as before but now {DATE} can be overwritten by the config.cfg file.
```
python3 -m bdfr download DIRECTORY --user USERNAME --submitted --sort hot --folder-scheme '{REDDITOR}' --file-scheme '{DATE}_{REDDITOR}_{TITLE}_{POSTID}'
```

PR #326, opened 2021-05-02
Add customisable time formatting

Serene-Arc:
Implements #321 

This should add a lot of flexibility if needed to the naming schemes, while still keeping ISO 8601 as the default and preferred method


aliparlakci (changes requested):
Can we have a documentation on this also?

Serene-Arc:
@aliparlakci woops, forgot that. Done now though

PR #349, opened 2021-05-06
Fix xml archiver encoding bug

aliparlakci:
XML archiver fails to archive [https://redd.it/n60t4c](https://redd.it/n60t4c) due to encoding issues. This issue is specific to Windows systems.


Serene-Arc (changes requested):
These tests should be consolidated. If you're only changing the single argument, then they can just all go in the same test.

aliparlakci:
I did this for it to be more verbose. I can merge them. Also, I think this tests can be included to `test_archiver.py` since the bug is related to _write_content_to_disk function and the function is not tested there.

Serene-Arc:
@aliparlakci sure, yeah that's probably a better place for them

PR #368, opened 2021-05-12
Bind socket to '0.0.0.0' rather than 'localhost' 

### LINKED ISSUE #366: [FEATURE] Consider binding OAuth socket to hostname rather than localhost
- [X] I am requesting a feature.
- [X] I am running the latest version of BDfR
- [X] I have read the [Opening an issue](../../README.md#configuration)
  
## Description
When running bdfr in a containerized environment, binding the socket used to receive the OAuth token info to 'localhost' causes complications and makes it impossible to expose the port externally. I would like to propose binding to the hostname or at least providing the ability to configure the hostname via an env var. 
I have tested using ```host = socket.gethostname()``` as the source of the host to bind to and have confirmed it works from within containers and on host machines.  I'd be more than happy to write, test, and PR this change if you think it would be helpful. Thanks!

BlipRanger:
This change enables authentication from within containers while still allowing for auth from host machines. 
Closes #366

### DIFF
--- a/bdfr/oauth2.py
+++ b/bdfr/oauth2.py
@@ -70,8 +70,8 @@ class OAuth2Authenticator:
     def receive_connection() -> socket.socket:
         server = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
         server.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
-        server.bind(('localhost', 7634))
-        logger.log(9, 'Server listening on localhost:7634')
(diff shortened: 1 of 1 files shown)


aliparlakci:
Could you add a user token as `REDDIT_TEST_TOKEN` secret to your fork repository and run the actions again?

BlipRanger:
> 
> 
> Could you add a user token as as secret to your fork repository and run the actions again? The secret name must be `REDDIT_TEST_TOKEN`

Ah yep, forgot to re-add that. Got it now.

PR #384, opened 2021-05-16
BREAKING: Change skip parameter names to be more consistent

aliparlakci:
### This PR:
- Changes `--skip` parameter to `--skip-format`
- Changes `--exclude-id` parameter to `--skip-id`
- Changes `--exclude-id-file` parameter to `--skip-id-file`


Serene-Arc (approved):
This seems like it's all in order, hopefully it's not going to be too confusing.

PR #417, opened 2021-05-21
Read files in chunks when calculating the existing hashes

### LINKED ISSUE #416: [BUG] Hash calculation takes up all the system memory for directories with large files.
- [x] I am requesting a feature.
- [x] I am running the latest version of BDfR
- [x] I have read the [Opening an issue](../../README.md#configuration)
  
## Description
The ability to limit downloading submissions at and over a certain file size, or another solution that would remedy the hash calculator taking up all memory and causing a system crash. Wasn't sure about labeling this bug or feature.

## Context
Because the youtube-dl fallback has been added, some of the videos bdfr downloads when I point it at a subreddit can be large in size (not ususally the case for say imgur/gfycat/redgifs). As such, when bdfr is run again (using `--search-existing ` and `--no-dupes` as usual), and is pointed at the same subreddit folder to get updates, it begins calculating hashes for these large files. In my example, there's only 15 files, all adding up to a 10gb directory.

The Ubuntu 20.04 laptop I use has 8gbs of RAM. When it begins calculating these hashes, it quickly takes up all the RAM, Ubuntu panics, all peripheral input is lost, and my laptop must be forcefully restarted by holding the power button. 

I analyzed my syslog at the time of the crash and it pointed to python3.9 with the correct PID as the culprit, after which no logs are written, and I forcefully restart. It shows out of memory (oom) trying to remedy the situation:

```
May 21 11:44:18 hostname kernel: [ 2671.186170] oom-kill:constraint=CONSTRAINT_NONE,nodemask=(null),cpuset=/,mems_allowed=0,global_oom,task_memcg=/user.slice/user-1000.slice/user@1000.service,task=python3.9,pid=7385,uid=1000
May 21 11:44:18 hostname kernel: [ 2671.186181] Out of memory: Killed process 7385 (python3.9) total-vm:1629364kB, anon-rss:784952kB, file-rss:0kB, shmem-rss:4kB, UID:1000 pgtables:2040kB oom_score_adj:0
May 21 11:44:18 hostname kernel: [ 2671.252607] oom_reaper: reaped process 7385 (python3.9), now anon-rss:0kB, file-rss:0kB, shmem-rss:4kB
```

Other than letting the user limit filesize, could there be a tweak to the hash calculation process to stop calculation if it reaches something like 80% (or some other percentage) of system memory? Maybe an option to disable the youtube fallback downloader?

I ran my script again and monitored the memory as it tried to calculate hashes for the specific folder. It went from ~20% to 90% and above as it calculated those hashes, as I suspected. [Here's a screenshot](https://i.imgur.com/NJNZhOK.png) of the memory reaching close to max capacity, and my attempt to ^c bdfr a couple times while it calculated the hashes (_calc_hash). It locked up shortly after. The keyboard interrupts didn't stop the memory usage from rising as fast as I thought it would 😛

aliparlakci:
Closes #416

### DIFF
--- a/bdfr/downloader.py
+++ b/bdfr/downloader.py
@@ -22,9 +22,15 @@ logger = logging.getLogger(__name__)
 
 
 def _calc_hash(existing_file: Path):
+    CHUNK_SIZE = 1024 * 1024
+    md5_hash = hashlib.md5()
     with open(existing_file, 'rb') as file:
-        file_hash = hashlib.md5(file.read()).hexdigest()
-        return existing_file, file_hash
+        chunk = file.read(CHUNK_SIZE)
+        while chunk:
+            md5_hash.update(chunk)
+            chunk = file.read(CHUNK_SIZE)
+    file_hash = md5_hash.hexdigest()
+    return existing_file, file_hash
 
 
 class RedditDownloader(RedditConnector):


Serene-Arc (approved):
Ah, well this is certainly an issue that I didn't anticipate when I wrote the code.  This seems to be a good fix for the issue though.

PR #427, opened 2021-05-25
Add a file blacklist to the direct downloader

### LINKED ISSUE #386: [BUG] Pages that has file extension in the URL causes bdfr to download the page source
## Description
Pages that has file extension in the URL causes bdfr to download the page source. In the below example, it should use youtube-dl to download the video in [https://www.pornhub.com/view_video.php?viewkey=ph5cc10ecef3aba](https://www.youtube.com/watch?v=dQw4w9WgXcQ). Instead it downloads a .php file with the page source. 

## Command
```
python -m bdfr download ~/Desktop/new --link "ndp8vc"
```

## Environment (please complete the following information):
 - OS: Windows 10
 - Python version: 3.9.4

## Logs
```
[2021-05-17 00:44:45,507 - bdfr.downloader - DEBUG] - Setting maximum download wait time to 120 seconds
[2021-05-17 00:44:45,508 - bdfr.downloader - DEBUG] - Setting datetime format string to ISO
[2021-05-17 00:44:45,509 - bdfr.downloader - Level 9] - Created download filter
[2021-05-17 00:44:45,509 - bdfr.downloader - Level 9] - Created time filter
[2021-05-17 00:44:45,509 - bdfr.downloader - Level 9] - Created sort filter
[2021-05-17 00:44:45,510 - bdfr.downloader - Level 9] - Create file name formatter
[2021-05-17 00:44:45,510 - bdfr.downloader - DEBUG] - Using unauthenticated Reddit instance
[2021-05-17 00:44:46,491 - bdfr.downloader - Level 9] - Created site authenticator
[2021-05-17 00:44:46,492 - bdfr.downloader - Level 9] - Retrieved subreddits
[2021-05-17 00:44:46,492 - bdfr.downloader - Level 9] - Retrieved multireddits
[2021-05-17 00:44:46,492 - bdfr.downloader - Level 9] - Retrieved user data
[2021-05-17 00:44:46,492 - bdfr.downloader - Level 9] - Retrieved submissions for given links
[2021-05-17 00:44:49,072 - bdfr.downloader - DEBUG] - Attempting to download submission ndp8vc
[2021-05-17 00:44:49,073 - bdfr.downloader - DEBUG] - Using Direct with url https://www.pornhub.com/view_video.php?viewkey=ph5cc10ecef3aba
[2021-05-17 00:44:50,434 - bdfr.downloader - DEBUG] - Written file to C:\Users\aliparlakci\Desktop\new\TSonFM\Secretsaucesource_Tgirl Fucks Pretty Teen Girl in the Ass_ndp8vc.php
[2021-05-17 00:44:50,434 - bdfr.downloader - DEBUG] - Hash added to master list: 9c75cd83ac653afedfc1e3ad4af1dbba
[2021-05-17 00:44:50,434 - bdfr.downloader - INFO] - Downloaded submission ndp8vc from TSonFM
[2021-05-17 00:44:50,435 - root - INFO] - Program complete

```

Serene-Arc:
Fixes #386

### DIFF
--- a/bdfr/site_downloaders/download_factory.py
+++ b/bdfr/site_downloaders/download_factory.py
@@ -61,12 +61,15 @@ class DownloadFactory:
     def is_web_resource(url: str) -> bool:
         web_extensions = (
             'asp',
+            'aspx',
             'cfm',
             'cfml',
             'css',
+            'htm',
             'html',
             'js',
             'php',
+            'php3',
             'xhtml',
         )
         if re.match(rf'(?i).*/.*\.({"|".join(web_extensions)})$', url):
--- a/bdfr/site_downloaders/download_factory.py
+++ b/bdfr/site_downloaders/download_factory.py
@@ -21,7 +21,7 @@ from bdfr.site_downloaders.youtube import Youtube
 class DownloadFactory:
     @staticmethod
     def pull_lever(url: str) -> Type[BaseDownloader]:
-        sanitised_url = DownloadFactory._sanitise_url(url)
+        sanitised_url = DownloadFactory.sanitise_url(url)
         if re.match(r'(i\.)?imgur.*\.gifv$', sanitised_url):
             return Imgur
         elif re.match(r'.*/.*\.\w{3,4}(\?[\w;&=]*)?$', sanitised_url):
@@ -49,7 +49,7 @@ class DownloadFactory:
                 f'No downloader module exists for url {url}')
 
     @staticmethod
-    def _sanitise_url(url: str) -> str:
+    def sanitise_url(url: str) -> str:
         beginning_regex = re.compile(r'\s*(www\.?)?')
         split_url = urllib.parse.urlsplit(url)
         split_url = split_url.netloc + split_url.path
--- a/bdfr/site_downloaders/download_factory.py
+++ b/bdfr/site_downloaders/download_factory.py
@@ -24,7 +24,8 @@ class DownloadFactory:
         sanitised_url = DownloadFactory.sanitise_url(url)
         if re.match(r'(i\.)?imgur.*\.gifv$', sanitised_url):
             return Imgur
-        elif re.match(r'.*/.*\.\w{3,4}(\?[\w;&=]*)?$', sanitised_url):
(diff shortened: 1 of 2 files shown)


bdfr/site_downloaders/download_factory.py
```diff
@@ -49,9 +50,26 @@ def pull_lever(url: str) -> Type[BaseDownloader]:
...
+        web_extensions = (
+            'asp',
+            'cfm',
+            'cfml',
+            'css',
+            'html',
+            'js',
+            'php',
+            'xhtml',
+        )
```
aliparlakci, line 71: `.php3`, `.htm`, `.aspx` are also common extensions.

aliparlakci (changes requested):

bdfr/site_downloaders/download_factory.py
Serene-Arc: Implemented with fef2fc8

aliparlakci:
Also, as can be seen in issue #417, there might be links ending with 4 characters after a dot which is not a file extension. For instance, https://www.reddit.com/user/lionessintherain/comments/d8qv4o/whoo_my_insta_is_back

Since it is very common to have dots in usernames and most websites has people's usernames in their path, I wonder if having a whitelist would be safer.

Serene-Arc:
I think the issue with that is that are so many filetypes out there and there's no conceivable way to include all of them reliably. There's just so many in so many categories like images, video, audio, archive, documents, and so on, each with their own quirks and common and uncommon types. If we go the whitelist route, I think that's inviting an endless stream of issues getting opened here saying, "Can you please add this single entry to the list of filetypes that I want to get?"

I think the current system is the better solution. Yes, there will occasionally be false-positives like the example that you gave, but they are much much rarer when compared to the false-negatives the alternative approach will cause IMO. Plus, these errors will only happen once if people are using the scripts to create a list of exclusionary submission IDs.

In my view, it is better for a downloader like the BDFR to err on the side of downloading rather than the opposite. If there's a choice, then it should try to retrieve as much as possible and let the user sort out the data after, rather than potentially missing something.

aliparlakci (approved):

Issue #429, opened 2021-05-26
[FEATURE] Allow specifying --user multiple times


OMEGARAZER:
- [x] I am requesting a feature.
- [x] I am running the latest version of BDfR
- [x] I have read the [Opening an issue](../../README.md#configuration)
  
## Description
Allow specifying --user multiple times.
#428

PR #432, opened 2021-05-27
Allow --user to be specified multiple times

Serene-Arc:
Implements #429

Thought I added this in when v2 was created but apparently not. This doesn't seem too complicated a change


OMEGARAZER:
Just thought I would drop a note here that I've been testing with this change locally and it's working well for me. Thanks very much for the implementation.

Issue #481, opened 2021-07-04
[FEATURE] Support inputting URLs/IDs from a text file


Nawor3565:
- [X] I am requesting a feature.
- [X] I am running the latest version of BDfR
- [X] I have read the [Opening an issue](../../README.md#configuration)
  
## Description
Add an option such as `--file my_urls.txt`, where `my_urls.txt` would be some text file with one URL/post ID per line. It would attempt to download each URL/ID, behaving just like the other modes such as `--subreddit`. This could be useful in many circumstances, but in my case, I would like to download everything in a list of all my saved posts which I got from a GDPR request (which bypasses the normal 1,000 post limit).

Thanks!

PR #482, opened 2021-07-05
Add ability to read IDs from files

Serene-Arc:
Implements #481 which people have been asking for for a while.

### DIFF
--- a/bdfr/__main__.py
+++ b/bdfr/__main__.py
@@ -6,9 +6,9 @@ import sys
 import click
 
 from bdfr.archiver import Archiver
+from bdfr.cloner import RedditCloner
 from bdfr.configuration import Configuration
 from bdfr.downloader import RedditDownloader
-from bdfr.cloner import RedditCloner
 
 logger = logging.getLogger()
 
@@ -17,6 +17,7 @@ _common_options = [
     click.option('--authenticate', is_flag=True, default=None),
     click.option('--config', type=str, default=None),
     click.option('--disable-module', multiple=True, default=None, type=str),
+    click.option('--include-id-file', multiple=True, default=None),
     click.option('--log', type=str, default=None),
     click.option('--saved', is_flag=True, default=None),
     click.option('--search', default=None, type=str),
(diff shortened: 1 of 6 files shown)


aliparlakci (changes requested):
I am not sure about this feature since it is same as running multiple commands back to back. Maybe we can make this specific to Reddit GDPR file templates.

Serene-Arc:
Kind of, this makes the BDFR more efficient over large series of IDs. There's a limit to how many you can fit on the command line before the OS errors out, and then you have to repeat the setup over and over again. That adds up over time. This makes it so that you can find IDs whichever way and then use it in a single command. Also makes it easier to use a variety of Bash tools and scripts like `uniq` and `diff` to sort and do more complex things with the lists.

The one thing I think we shouldn't do is tie it to the GPDR templates. That's narrowing the focus of a broadly useful feature to a single use case when it should be trivial to extract the data from the CSV in the first place. I feel that the BDFR should be as data-agnostic as possible. It shouldn't matter where we got the list of IDs, just that there is a list. Then the BDFR can download them after whatever preprocessing the user desires has been done.

Serene-Arc:
@aliparlakci are we good to merge this PR?

bdfr/__main__.py
```diff
@@ -17,6 +17,7 @@
     click.option('--authenticate', is_flag=True, default=None),
     click.option('--config', type=str, default=None),
     click.option('--disable-module', multiple=True, default=None, type=str),
+    click.option('--include-id-file', multiple=True, default=None),
```
aliparlakci, line 20: `--from-id-file` more suitable

aliparlakci (changes requested):

aliparlakci:
I have no other comment, it can be merged after the parameter name change.

bdfr/__main__.py
Serene-Arc: I think the current name is better, it provides an opposite to the option `--exclude-id-file`. Since the latter already exists, I think the same convention for the former would be less confusing.

PR #503, opened 2021-07-29
Implement callbacks for downloading

Serene-Arc:
Implemented callbacks so now when there's a video downloaded, it'll download in the same place as everything else. Also means that skipping videos will properly work now.

### DIFF
--- a/bdfr/resource.py
+++ b/bdfr/resource.py
@@ -6,7 +6,6 @@ import logging
 import re
 import time
 import urllib.parse
-from collections import namedtuple
 from typing import Callable, Optional
 
 import _hashlib
@@ -33,8 +32,12 @@ class Resource:
     def retry_download(url: str) -> Callable:
         max_wait_time = 300
 
-        def http_download() -> Optional[bytes]:
+        def http_download(download_parameters: dict) -> Optional[bytes]:
             current_wait_time = 60
+            if 'max_wait_time' in download_parameters:
+                max_wait_time = download_parameters['max_wait_time']
+            else:
+                max_wait_time = 300
             while True:
                 try:
                     response = requests.get(url)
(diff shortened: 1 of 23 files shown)


bdfr/resource.py
```diff
@@ -18,40 +18,52 @@
...
-        except (requests.exceptions.ConnectionError, requests.exceptions.ChunkedEncodingError) as e:
-            logger.warning(f'Error occured downloading from {url}, waiting {current_wait_time} seconds: {e}')
-            time.sleep(current_wait_time)
-            if current_wait_time < max_wait_time:
-                current_wait_time += 60
-                return Resource.retry_download(url, max_wait_time, current_wait_time)
+    def retry_download(url: str) -> Callable:
+        max_wait_time = 300
+
+        def http_download(download_parameters: dict) -> Optional[bytes]:
```
aliparlakci, line 35: Why not `def http_download(*args, **kwargs)`?

aliparlakci:
I looked at this and it seems reasonable to me. I like FP :)

> it'll download in the same place as everything else. Also means that skipping videos will properly work now.

But can you elaborate more on how this PR improves the areas you mentioned above (and also the ones you didn't mentioned but thought of) just so we are on the same page?

Serene-Arc:
Oops, misclick. Didn't mean to close it.

What's FP?

You know how videos were downloaded at the "find links" stage? This will fix that. It used to be that everything else downloaded in the download stage, but videos had a different flow due to using youtube-dl. So this will make it so that all downloads occur at the same place.

The old way had it downloading before it had a chance to skip any submissions so even if you wanted to skip videos, it would be downloaded anyway. Big waste of bandwidth and time, so this should fix that.

aliparlakci (approved):
:shipit:

Issue #508, opened 2021-08-13
[FEATURE] Unsave Saved Posts


joeswanson1337:
This used to be a feature (suggested in #70 and implemented in #103) in 1.8, but it seems that it was removed. 
This seems like a useful feature as it allows you to tidy up the saved page and makes it easier to avoid the ~1000 saved post cap.

kc2rxo:
+1 to this feature.

Issue #70, opened 2019-11-08
Feature Request: Option to unsave items as they're downloaded


ErikLentz:
To keep my Reddit saves section clean, and to keep clear what has and hasn't been downloaded, I would love if it bulk-downloader could unsave items as it downloads them.

Draxxx:
I'll second this. That'd be a great feature.

DigiOhhh:
Third here!

aliparlakci:
@Draxxx @DigiOhhh @ErikLentz 

Feature is live on the 1.8 version. Use the option --unsave

DigiOhhh:
This is simply amazing, thank you so much!
Would it be possible to do the same thing with upvoted posts?

PR #509, opened 2021-08-14
feat: unsave posts after downloading

kc2rxo:
This commit adds the feature request in #508. The feature was originally requested in #70, which was implemented in #103 before being removed in what looks like the v2 upgrade.

### DIFF
--- a/bdfr/__main__.py
+++ b/bdfr/__main__.py
@@ -46,6 +46,7 @@ _downloader_options = [
     click.option('--skip', default=None, multiple=True),
     click.option('--skip-domain', default=None, multiple=True),
     click.option('--skip-subreddit', default=None, multiple=True),
+    click.option('--unsave', is_flag=True, default=None)
 ]
 
 _archiver_options = [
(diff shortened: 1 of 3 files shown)


Serene-Arc:
I'm against this feature as a matter of principle for the BDFR. I do not believe that a downloader should modify remote content in this way, especially since it's the only place that modification takes place. Feels out of step with the rest of the function of the BDFR, especially given that for many people actually downloading their full saved post list requires a GPDR request.

I think that this feature would be best suited towards a daughter project that's been talked about for some time, with an interactive interface while using this project as a backend. There's no timeframe on that though and I haven't given it any thought.

Given that my opposition to this is entirely based on opinion and principled views of the project and has zero technical basis, I'll ping @aliparlakci and let him decide since it is his tool.

As far as this PR is concerned from the perspective of the lead maintainer, it needs some tests, primarily integration tests, to confirm that this is in fact working. Once those are written, if @aliparlakci wishes to merge then it should be good to go.

aliparlakci:
This can easily be done with a simple script and the log_output. No need to implement it internally. I will add the script to /script folder in the project as soon as possible.

Thank you for your work but this won't be merged.

Issue #513, opened 2021-09-03
[FEATURE] Skip user


madhatr:
- [x] I am requesting a feature.
- [x] I am running the latest version of BDfR
- [x] I have read the [Opening an issue](../../README.md#configuration)
  
## Description
I would like the ability to skip certain users content when downloading a subbreddit.  I looked through the settings and configurations but didn't see it.  I saw a few other skip for ID(?) and subreddits.

Edit: Added more info, want to skip users while downloading subreddits.

madhatr:
Oh sweet.  Thank you!

PR #546, opened 2021-10-31
Add --ignore to ignore user

jrwren:


jrwren:
for #513

Serene-Arc (changes requested):
Hi, few things:

- The option you added to the documentation is `--ignore`, not `--ignore-user`
- The if statement shouldn't be in the cloner, it should be in the downloader class. The cloner simply inherits from both of them
- If you could write tests for the relevant sets of integrations (the cloner, downloader, and archiver) that would be great. The BDFR needs to have every change tested

Serene-Arc (changes requested):
Hey, thanks for fixing the documentation. This test doesn't pass, in fact it doesn't run at all. I more meant integration tests, although it's not bad to have this one as well.

PR #551, opened 2021-11-10
Stop videos from being downloaded as images

OMEGARAZER:
Erroneous .gifv extensions such as .giff .gifff or .gift resolve to a static image and are downloaded by the direct downloader. (ex: https://i.imgur.com/OGeVuAe.giff  )


Serene-Arc (changes requested):
Hi, thanks for contributing :) For this could you please add tests showing that these extensions are now caught and resolve to the specific resource i.e. add tests for urls that would fail now and would succeed after this change.

OMEGARAZER:
I've added to the test cases, if I missed anything please let me know.

Serene-Arc (approved):

PR #579, opened 2022-01-05
Add equivalent PowerShell scripts

### LINKED ISSUE #414: [DOCS] Add equivalent powershell scripts
## Description
Add equivalent powershell scripts to [`/master/scripts`](https://github.com/aliparlakci/bulk-downloader-for-reddit/tree/master/scripts)

Related to #412

Thayol:
Resolves #414 

The PowerShell scripts should be equivalent of the Bash scripts. Could not test every situation, a large log file with every possibility would be welcome!

Also added a `.gitattributes` file to keep the `CRLF` line endings for Windows execution.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+# Declare files that will always have CRLF line endings on checkout.
+*.ps1    text  eol=crlf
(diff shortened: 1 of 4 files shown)


Serene-Arc:
The easiest way to get the log files for every test are to use the ones in the bash tests, since they're the same format. The strings are in the code if you really want to use them too, you can substitute real-world values in pretty easily.

Thayol:
I've processed the test log files with PowerShell and it seems that even the Bash scripts produce the wrong result for [failed_write_error.txt](https://github.com/aliparlakci/bulk-downloader-for-reddit/blob/e4fcacfd4fc3d8151e98cf935034012f7132cc8f/scripts/tests/example_logfiles/failed_write_error.txt)

Expected result: `nnboza`
Actual result: `submissio`

Currently, the Bash and PowerShell scripts behave identically. Should I create a new issue with the "write error" bug or should I include the solution in this PR?

Serene-Arc:
Ah well if the bash scripts cause an error currently, open a new bug report for that. You or I can write a PR, but keep the powershell script separate. You can always rebase if needed

Thayol:
So the question is whether I should keep things like this and make the PowerShell scripts mirror what the Bash scripts do or should I commit (in this PR) the PS script that produces the "correct" result?

Asking because I can't make my next pull request that would "fix" the PS scripts too if the PS scripts don't exist.

Serene-Arc:
Make the PS scripts do the right thing, don't knowingly commit a broken script. I'll check and merge the other PR

Thayol:
Now it is aligned with the test data, everything works as intended.

PR #601, opened 2022-02-18
Add support for Redgifs images and galleries

Serene-Arc:


Soulsuck24:
Thanks for the quick turnaround on this. Testing it it seems to download the images as mp4 files. Is that correct for the way it's set up or it is possibly doing things weird on my end?

Serene-Arc:
@Soulsuck24 oops that's an oversight on my part, I'll get on that. If you could open another bug report, please do, otherwise I will

PR #626, opened 2022-04-18
Use stdout

chapmanjacobd:
It's kind of a strange interface to create a file... that should be left up to the user. 

With stdout you can use process substitution to interact with bdfr, for example in BASH:

```
python -m bdfr download bdfr/ --include-id-file <(~/bin/bdfr-failed.sh ~/.config/bdfr/log_output.txt*) 
```

or fish shell:

```
python -m bdfr download bdfr/ --include-id-file (~/bin/bdfr-failed.sh ~/.config/bdfr/log_output.txt* | psub) 
```

### DIFF
--- a/scripts/extract_failed_ids.sh
+++ b/scripts/extract_failed_ids.sh
@@ -7,17 +7,10 @@ else
     exit 1
 fi
 
-if [ -n "$2" ]; then
-    output="$2"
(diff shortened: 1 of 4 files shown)


Serene-Arc:
Hi, seems fine to me, but please update the Bash tests for this change.

PR #632, opened 2022-04-25
Feature/v.reddit

BlipRanger:
Adds direct support for v.redd.it video links. Added a few tests. Passes all tests but 4 unrelated. ones


chapmanjacobd:
this is kinda dangerous, no? https://github.com/aliparlakci/bulk-downloader-for-reddit/pull/632/commits/99fe3312a4f343fb45b6d55cf21e54bb10f9dd33

```
ydl.download([self.post.url])
```

Instead of writing hundreds of lines of custom code re-implementing a yt-dlp (or youtube-dl) wrapper, it might make more sense for all videos to be passed to yt-dlp (or youtube-dl). It would be a lot more maintainable that way. yt-dlp (or youtube-dl) supports [a lot of sites](https://github.com/yt-dlp/yt-dlp/blob/master/supportedsites.md) and it would be very tedious and redundant to write a wrapper for each one...

For images as well there is `gallery-dl` which has good support for many sites. I'm not sure what integration would look like but it might be interesting to find a way to support downloading from many sites without duplicating the work that people have done before

chapmanjacobd:
It's actually possible to use ytdl within gallerydl

https://github.com/mikf/gallery-dl/blob/master/gallery_dl/extractor/ytdl.py

I wonder if bdfr could use a similar strategy or maybe improving the gallerydl reddit extractor is a better strategy in the long run I'm not sure

BlipRanger:
> > Bind socket to '0.0.0.0' rather than 'localhost
> > [99fe331](https://github.com/aliparlakci/bulk-downloader-for-reddit/commit/99fe3312a4f343fb45b6d55cf21e54bb10f9dd33)
> 
> this is kinda dangerous, no?
> 


Oh, that change actually isn't changing anything if you look at the existing code; that's already how they're doing it. I believe that is necessary for being able to do the oauth within docker though.

BlipRanger:
> It's actually possible to use ytdl within gallerydl
> 
> https://github.com/mikf/gallery-dl/blob/master/gallery_dl/extractor/ytdl.py
> 
> I wonder if bdfr could use a similar strategy or maybe improving the gallerydl reddit extractor is a better strategy in the long run I'm not sure

Definitely could be an improvement, though this PR really is just a quick patch for the v.reddit stuff. I'd definitely like to see another that explores gallery-dl though.

PR #668, opened 2022-09-26
Fix Vidble/Erome tests

Soulsuck24:
Fixes broken link in Vidble test.

these watch?= seem short lived, not common and tricky to track down, may need to consider removing it from future tests. In two days I've needed to change the link to test 4 times as they disappear. Hopefully this one hangs around long enough to work though.


Soulsuck24:
Also managed to figure out what was happening with the Erome tests. If I'm not mistaken that's all the primarily "Adult" downloaders taken care of for you @Serene-Arc

Serene-Arc:
This one is out of date too. I've updated the regex in the Erome class already and the Vidble test case has been removed due to the source being deleted.

PR #669, opened 2022-09-26
Possible fix for YTDL based tests.

OMEGARAZER:
Seems the file downloads the same but now has a different hash. Please verify that it's not just on my end but may just be YTDL downloading with different settings/quality.


Serene-Arc:
I took a different route and updated the minimum yt-dlp version. It was almost a year out of date and with the new version, support for more sites has been added. The new hashes reflect this version so this PR is kinda moot, sorry.

OMEGARAZER:
Sorry, apparently I missed your changes before I got these in. My bad.

Serene-Arc:
All good. I push directly to development when not responding to a PR or issue, so it's good to check those. Mainly these days it's just test updates, so you're the first to have trouble with it. Thanks though

PR #671, opened 2022-09-28
Imgur edge case coverage

OMEGARAZER:
Covers edge case of additional arguments on extension.

Also removed duplicate or redundant tests.

### DIFF
--- a/bdfr/site_downloaders/imgur.py
+++ b/bdfr/site_downloaders/imgur.py
@@ -78,7 +78,7 @@ class Imgur(BaseDownloader):
 
     @staticmethod
     def _validate_extension(extension_suffix: str) -> str:
-        extension_suffix = extension_suffix.strip('?1')
+        extension_suffix = re.sub(r'\?.*', '', extension_suffix)
         possible_extensions = ('.jpg', '.png', '.mp4', '.gif')
         selection = [ext for ext in possible_extensions if ext == extension_suffix]
         if len(selection) == 1:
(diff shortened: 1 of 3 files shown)


bdfr/site_downloaders/imgur.py
```diff
@@ -78,7 +78,7 @@ def _get_data(link: str) -> dict:
 
     @staticmethod
     def _validate_extension(extension_suffix: str) -> str:
-        extension_suffix = extension_suffix.strip('?1')
+        extension_suffix = re.sub(r'\?.*', '', extension_suffix)
```
Serene-Arc, line 81: Please add tests that show why this is necessary e.g. a test that fails before this change and succeeds afterwards.

Serene-Arc (commented):
Why have you removed working tests from here?

OMEGARAZER:
Duplicate aside, with the factory sending anything with the Imgur domain to its own downloader rather than checking the extension on the link before deciding to send to Imgur or Direct the tests don't really serve any purpose outside of padding test count.

One link there is related to the other changes but that test would take place separately and currently does not have a test. I will have to dig into pytest to figure out how to properly add one for that.

bdfr/site_downloaders/imgur.py
OMEGARAZER: Outside what I mentioned in my other comment this covers ?2, ?3 and any others that may happen in the future. I'll go back and dig out the on that triggered this to go with the ?1 example and work on a test with them.

OMEGARAZER: I've added test cases that should hopefully satisfy this case, Not sure if there's a better way to test it but they verify they work currently.

PR #693, opened 2022-11-08
Fix test for deleted user on post

OMEGARAZER:
tested post now showing deleted as user causing tests to fail. Updated to working post.


OMEGARAZER:
@Serene-Arc Can you check the second commits test to make sure I haven't messed up what you were going for.

I removed the link that was in both from the first as it just trew the except that the second test was catching. If it was in there for another reason I appologize.

PR #698, opened 2022-11-26
Update tests

OMEGARAZER:
Suspended user in two tests.

Updated hashes and yt-dlp version.

Removed success check on known failure.


Serene-Arc:
You removed the two new tests from the cloner integration, any reason for that?

OMEGARAZER:
They're checking for successful clone of known 403 errors. They won't pass.

Issue #472, opened 2021-06-22
[BUG] Various downloader issues


Soulsuck24:
- [x] I am reporting a bug.
- [x] I am running the latest version of BDfR
- [x] I have read the [Opening an issue](https://github.com/aliparlakci/bulk-downloader-for-reddit/blob/master/docs/CONTRIBUTING.md#opening-an-issue)
  
## Description
Sorry I didn't break these into multiple issues but some downloaders only have 1-2 and the report would be substantially similar.
posts are from various users over various dates but all fail for various reasons. No direct error messages in logs, just straight from "attempting to download..." to errors in Log section.

NSFW posts as an FYI.
Imgur ID's: ltb0wg (? with no number on end of link), mfd4o6/mfd61f/mfd8zz (capitalized gifv)
Gallery ID's: nxyahw/o4kune/o6omj4 (Says no images in gallery but available through web)
Redgifs ID's: ljpqo8 (seems to think it's a gyfcat link when it's not), mfxjbc/mrcu38/m0ouga/m0opyt/mf4wag/nndlss/nqb80o/no0hds (All give 404 but are available through web)

## Environment (please complete the following information):
 - OS: Ubuntu 21.04
 - Python version: 3.9.5

## Logs
```
Site Imgur failed to download submission ltb0wg: Could not read page source from https://i.imgur.com/ywSyILa.gifv?
Failed to download resource https://imgur.com/ubYwpbk.GIFV in submission mfd4o6 with downloader Direct: Unrecoverable error requesting resource: HTTP Code 404
Failed to download resource https://imgur.com/5b3TuEy.GIFV in submission mfd61f with downloader Direct: Unrecoverable error requesting resource: HTTP Code 404
Failed to download resource https://imgur.com/hdlttfd.GIFV in submission mfd8zz with downloader Direct: Unrecoverable error requesting resource: HTTP Code 404

Site Gallery failed to download submission nxyahw: No images found in Reddit gallery
Site Gallery failed to download submission o4kune: No images found in Reddit gallery
Site Gallery failed to download submission o6omj4: No images found in Reddit gallery

Site Redgifs failed to download submission ljpqo8: Server responded with 404 to https://api.redgifs.com/v1/gfycats/BlushingPortlyBarnswallow
[...]

PR #699, opened 2022-11-27
Redgifs improvements

Soulsuck24:
Add check to verify token was received.
Update headers sent to content API.

Add availability check for videos to resolve last part of #472 where only SD version is available.


Serene-Arc:
Thanks, it's great someone else is willing to look at all the porn to fix these modules instead of me.

PR #705, opened 2022-12-01
Pylance typing

OMEGARAZER:
Fix Pylance warnings for typing


Serene-Arc:
Actually if we're going down this route, I'd prefer we use the newer typing method of doing unions, namely `str | bytes` and the like.

OMEGARAZER:
@Serene-Arc Just noticedwe'll need to update min version to 3.10 with that formatting

Serene-Arc:
It's been reverted. Much as I would like to switch to 3.10 we should probably wait for Debian at the very least. That puts the upgrade time at mid next year at this rate. For now, we should stick to 3.9 compatible typing. If you could do the PR without my change again, that would be great, you were right the first time.

PR #713, opened 2022-12-11
Add soft fail on 5xx Prawcore errors.

OMEGARAZER:
Adds except to stop 5xx errors killing run when going through user/submission lists.

Wasn't able to figure out how to write a test for this as it's an intermittent issue and not one we can knowingly get like 4xx errors. Been running with it for a few days and it's been working well though.

Also reduces some log spam by only printing the user being fetched rather than the whole list.

### DIFF
--- a/bdfr/archiver.py
+++ b/bdfr/archiver.py
@@ -4,6 +4,7 @@
 import json
 import logging
 import re
+from time import sleep
 from typing import Iterator, Union
 
 import dict2xml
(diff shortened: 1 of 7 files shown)


Serene-Arc:
You can use the patch function from the unit testing function to manually through a praw exception for the test to simulate the server messing up.

OMEGARAZER:
> You can use the patch function from the unit testing function to manually through a praw exception for the test to simulate the server messing up.

I can't seem to get patch to work for me in the way I need to for this (or at least how I'm trying?). I can get it to raise the exception directly but not indirectly where it shows these catching the exception. So any help you can give would be much appreciated.

Serene-Arc:
Sure I'll take a look and add the tests to this PR.

Serene-Arc:
Oh cool did you get them to work?

OMEGARAZER:
It appears so, If I remove my original patches the tests fail, killing BDFR but tests pass with them.

PR #720, opened 2022-12-17
Consolidate to pyproject

OMEGARAZER:
Consolidates configs to pyproject.toml and updates workflows accordingly as well as sets sane minimums for dev requirements.

adds version check to main script.


Serene-Arc:
Cool, great work.

Serene-Arc:
So does this force an exit if the version isn't up to date? I think it should show the message but still allow the tool to be used.

OMEGARAZER:
If you mean the exit in the version callback then no, it doesn't stop anything from working. That's just there so when you run version it doesn't say you're missing the directory. It also only hits that when you use --version to call it.

Serene-Arc:
Ah, right. All good then

PR #722, opened 2022-12-18
Cleanup tests

OMEGARAZER:
Cleans up test args on new tests.

Add log path to default config test so as not to mangle default log outside of tests.

Match setup functions to archive/clone.

Remove testing marker that was commited in error.


Serene-Arc:
Still that damn Mac error. Changes seem fine though

OMEGARAZER:
Yeah, if it was something consistent it could be masked but it seems to be different every time.

PR #729, opened 2022-12-20
Standardize shebang and coding declaration

OMEGARAZER:
Standardizes shebang and coding declarations.

Coding matches what's used by install tools such as pip(x).

Removes a few init files that were not needed.


OMEGARAZER:
For my future information do you have examples? Didn't run into any issues and messed around with it before submitting this, figured it was related to the old pbr build.

Serene-Arc:
What you did is create a namespace package [instead of a normal one](https://docs.python.org/3/reference/import.html). Normal packages have an `__init__.py` file in every directory, while namespace ones don't and can be split up in multiple areas of the system's PATH. It still works because they are mutually compatible to some degree but there's no reason to change it when the BDFR doesn't utilise multi-directory structures.

OMEGARAZER:
I meant examples of it being temperamental, I was just curious though, it's not an issue.

PR #708, opened 2022-12-01
2.6.1

OMEGARAZER:
With the number of fixes and changes (There's a non-working module in 2.6.0), it's probably a good idea to get a release out.

It also would leave a good starting point to work toward #703 and any other automatic linting/formatting.

PR #734, opened 2022-12-27
Add ability to retest

OMEGARAZER:
### ***Ability to retest won't be available till on the main branch

Adds ability to re-run tests if required.

Ignores if only markdown files are changed as they are handled by tox and weren't touched by these tests.

Adds basic Refurb settings to pyproject. Not adding as dev requirement as it's a nice to run but by no means needed.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -3,11 +3,18 @@ name: Python Test
 on:
   push:
     branches: [ master, development ]
+    paths-ignore:
+      - "*.md"
   pull_request:
     branches: [ master, development ]
+    paths-ignore:
+      - "*.md"
+  issue_comment:
+    types: [ created ]
 
 jobs:
   test:
+    if: contains(fromJson('["push", "pull_request"]'), github.event_name) || ( ${{ github.event.issue.pull_request }} && contains(github.event.comment.body, '>retest-bdfr') )
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
--- a/README.md
+++ b/README.md
@@ -1,8 +1,12 @@
 # Bulk Downloader for Reddit
 
-[![PyPI version](https://img.shields.io/pypi/v/bdfr.svg)](https://pypi.python.org/pypi/bdfr)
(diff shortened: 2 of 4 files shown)


Serene-Arc:
What do you mean rerun tests?

Serene-Arc:
Also please separate out the refurb stuff to another PR, we can add it to the pre-commit stuff as well, and move towards integrating it into the GH checks.

OMEGARAZER:
> What do you mean rerun tests?

Like it says on the tin, lets the tests be re-run in situations like [this](https://github.com/aliparlakci/bulk-downloader-for-reddit/issues/708#issuecomment-1335918020) where the tests are failing outside our control. Rather than being forced to make another change and hope the tests pass as they should a new comment on the PR can be made with `>retest-bdfr` which should re-ren the tests without needing to make any changes.

> Also please separate out the refurb stuff to another PR, we can add it to the pre-commit stuff as well, and move towards integrating it into the GH checks.

I figured it wasn't worth another PR for 4 lines but if it's going to be integrated more I can move it out.

Serene-Arc:
The thing is, I don't like this approach. I want to tests to be fixed rather than having this fix in place and complicating the testing process. If there's no way to fix them, then maybe this can be merged but there should be at least an attempt to fix the failing tests so they do so regularly or pass regularly.

OMEGARAZER:
> I want to tests to be fixed rather than having this fix in place and complicating the testing process.

This doesn't change or alter any tests so I wouldn't call this a fix by any means and it never intended to be.

> If there's no way to fix them

My point with this though is that in the situation like the one I pointed out there was nothing _to_ fix.

Things like all the tests passing in my fork when I publish a change but then a test (seems mac mostly) will fail with 5xx/4xx errors in the PR on completely untouched sections.

Was just an attempt to lessen situations where tests are known good but failing outside of BDFR.

Serene-Arc:
You don't have to close it, I just want to double check for reasons why it's failing on MacOS and see if there's any reasons for the failures first.

PR #735, opened 2022-12-27
Adds Refurb to testing

OMEGARAZER:
Adds Refurb to testing/pre-commit

EDIT: Appears I forgot that refurb requires 3.10 to run but will still check for code down to 3.6.


OMEGARAZER:
There are also some updates that can be done to the flake8 lint. Would you like them separate or would adding them with this be okay? they don't change anything in the raw testing, just consolidates the config for it.

PR #736, opened 2022-12-31
Consolidate flake8 settings

OMEGARAZER:
Consolidates sane flake8 settings to pyproject with the Flake8-pyproject plugin.

Does not change logic of test workflow but allows base settings to live in pyproject for anyone using flake8 as an external linter (e.g. vscode)

Also fixes some flake8 errors that were not being picked up by current testing, mostly unused imports.


Serene-Arc:
Perhaps add flake8 to the precommit configuration?

PR #741, opened 2023-01-01
2.6.2

OMEGARAZER:
Probably wouldn't hurt to get a release out with #740 having hit as well as all the other fixes and updates that have gone in since 2.6.1


OMEGARAZER:
Lovely mac tests...

_**facedesk**_

PR #747, opened 2023-01-05
Add hash-file option

OMEGARAZER:
Adds --hash-file option to seed the master_hash_list to assist the --no-dupes option.

Simply adds the option to be used, is not required and functions similarly to already existing --exclude-id-file.


Serene-Arc:
I'm still really opposed to adding this kind of thing to the BDFR, I don't think we should be doing anything with hashes. Adding this option seems like it would be encouraging this system to be expanded in the future. It's a holdover from v1 more than anything. 

One of my tenets from this program is that the files on disk are the only source of information, the master record. Using a list of pre-calculated hashes to represent what's on disk is a contradiction of that.

OMEGARAZER:
What expansion past this are you worried about?

If it were required to be used, I'd agree that it shouldn't be added lightly, but it's an optional upgrade to a function already contained in the script.

By your tenet that the files on disk should be the only source of info and not use hashes the exclude-id-file/hard-link/no-dupes options should all be removed is that correct?

Serene-Arc:
`--no-dupes` is only about the files downloaded in that run, not the files on disk. The exclude id options aren't the same thing, since they don't actually depend on the files on disk. The hard-link option doesn't change that either.

None of these options are the same as the one you're proposing. The IDs in the `--exclude-id` options cannot be rendered out of date, but the same is not true for file hashes. Any change to metadata for a file will instantly render the hash list obsolete, so relying on it to decide what to download cannot be a reliable method.

I still think anything to do with hashes shouldn't really be the BDFR's business. It's not a file manager, it's supposed to ject download from Reddit. Managing and filtering IDs is part of that, but trying to identify files and hashes on the fly seems a bit outside of scope, if that makes sense?

I'm not sure, but when I wrote these features, it was mostly just a porting features from v1 and the BDFR has changed a lot since then.

OMEGARAZER:
> `--no-dupes` is only about the files downloaded in that run, not the files on disk.
> The hard-link option doesn't change that either.

But both use hashes to do so.

> The exclude id options aren't the same thing, since they don't actually depend on the files on disk.

But are an external source of information outside of the files on disk.
 
> Any change to metadata for a file will instantly render the hash list obsolete, so relying on it to decide what to download cannot be a reliable method.

I'm not sure I understand what you mean by this. If the hash is dumped from the master hash list after a run and then the metadata changes locally it's not the new hash that I'm worried about ignoring, it's the old one. If the metadata changes upstream on the poster or Reddit side then it's no different than a repost when excluding the old post id, it gets re-downloaded. Both are obsoleted by upstream changes.

> I'm not sure, but when I wrote these features, it was mostly just a porting features from v1 and the BDFR has changed a lot since then.

Sure, but that shouldn't stop features that are already there from being more efficient at what they do. In the week or two I've been using this locally I've had *maybe* 1-2% of duplicate files that I was getting before and the ones that do are because of things mentioned above like edits in the repost or metadata changes which nothing else would have caught anyway.

Serene-Arc:
Excluding IDs are not the same as excluding hashes. IDs are, in a sense, independent of files since it happens at the stage of gathering Reddit sources, not downloading files. There is a distinction. That's why we allow including subreddits, even though they are not 'files'.

Hard-links are not the same as hashes? I'm not sure what your point is here. Yes `--no-dupes` uses hashes. I can't get rid of that without making it v3 so it's staying in. Same with the hardlinks, which was my mistaken attempt to approve a bad set of features.

Look, the bottom line is that this is the exact reason I *don't* want these features expanded upon. I expanded `--no-dupes` with the ability to make hard-links instead. Now you're pointing to `--no-dupes` *and* the hard links option to justify this option. The next PR will reference no dupes, the hard links, and this feature for why it should be included. They'll just keeping being expanded.

> I'm not sure I understand what you mean by this. If the hash is dumped from the master hash list after a run and then the metadata changes locally it's not the new hash that I'm worried about ignoring, it's the old one. If the metadata changes upstream on the poster or Reddit side then it's no different than a repost when excluding the old post id, it gets re-downloaded. Both are obsoleted by upstream changes.

Then what's the point of this feature? Like, what problem are you trying to solve? You're not minimising downloads; you need to download the data to hash it. Are you trying to minimise the number of files on disk? My position is that other tools like rdfind are better suited and the BDFR isn't going to attempt that. Are you trying to keep a list of hashes downloaded for the sake of it? Not the BDFR's job either.

If you can give a problem that this feature solves, that isn't just doing rdfind but worse, please, I'd love to hear it. But right now this seems to just expand a system that I've repeatedly said shouldn't be used.

OMEGARAZER:
> Hard-links are not the same as hashes?

I never said they were?

> Are you trying to minimise the number of files on disk? My position is that other tools like rdfind are better suited ... just doing rdfind but worse

You don't do something worse by not needing to use it in the first place. Almost completely removing the need for another tool for 28 lines, plus less thrash and wear on disks seems like a good trade-off to me.

Serene-Arc:
Except you're not eliminating the need to use it. If the hashes are rendered out of date, if there's a single run without the option enabled, if the the files are transferred between systems, all of which are scenarios the BDFR is explicitly designed to support, would mean that this option is useless and a tool such as rdfind would have to be able to used in the end anyway.

Again, what problem are you trying to solve here? Is it minimising wear and tear on disks? Because there's a lot that can be done for that which isn't in the BDFR anyway. If it's that much of a concern, don't download a ton of small files. You don't eliminate the need for another tool, you expand a system that the BDFR does poorly and which is borderline out of scope. 

If this solves a problem that isn't just trying to limit duplicates written to disk in this limited set of circumstances, please describe it. Otherwise I'll close the PR.

Botts85:
> Except you're not eliminating the need to use it. If the hashes are rendered out of date, if there's a single run without the option enabled, if the the files are transferred between systems, all of which are scenarios the BDFR is explicitly designed to support, would mean that this option is useless and a tool such as rdfind would have to be able to used in the end anyway.
> 
> Again, what problem are you trying to solve here? Is it minimising wear and tear on disks? Because there's a lot that can be done for that which isn't in the BDFR anyway. If it's that much of a concern, don't download a ton of small files. You don't eliminate the need for another tool, you expand a system that the BDFR does poorly and which is borderline out of scope.
> 
> If this solves a problem that isn't just trying to limit duplicates written to disk in this limited set of circumstances, please describe it. Otherwise I'll close the PR.

@Serene-Arc I had been using a solution similar to @OMEGARAZER's as a workaround to an issue with symlinks and --search-existing since before he submitted the pull request.  

At present, if a user is using a tool like rdfind or jdupes to locate and replace duplicates with symlinks then runs --search-existing it can cause issues with the MD5 hashing since it is failing when it hits a symlink. 

That could be addressed within in _calc_hash by following the symlink or just ignoring symlinks which would work on macOS and linux, but I am not sure about how Windows would handle that.

Serene-Arc:
Your comment just proves why I don't want this merged. I want to discourage the use of `--search-existing` as much as possible. I'm not going to add features that make using that option easier, it shouldn't be used at all. 

Don't use `--search-existing` at all, just use rdfind or jdupes in the first place. This is a solution to a problem that already has one.

PR #769, opened 2023-02-04
Gfycat API

OMEGARAZER:
Moves Gfycat to use API via site access key.

Adds cachetools as dependency to reuse API keys for Gfycat/Redgifs at 95% of their TTL. Include tests to verify caching.

Updates versions of requests/yt-dlp/black/isort/pytest.

Added default timeout to requests calls.

Adds validate-pyproject and blacken-docs to pre-commit as well as updates versions.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -2,13 +2,18 @@
 # See https://pre-commit.com/hooks.html for more hooks
 
 repos:
+  - repo: https://github.com/abravalheri/validate-pyproject
+    rev: v0.12.1
+    hooks:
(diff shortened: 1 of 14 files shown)


OMEGARAZER:
Was there anything else needed for this or any input?

@aliparlakci @Serene-Arc

Serene-Arc:
@OMEGARAZER in the future, I'd prefer that you do these changes in separate PRs. Each of the lines in your summary should be a separate PR as they are unrelated. Moving forwards I won't merge PRs like this because they really confuse the process and git history. They also make reviewing PRs a lot harder, fyi.

PR #780, opened 2023-02-18
Quote cleanup

OMEGARAZER:
Cleanup of string quotes to prioritize outer double quotes and use bugbear B907 to clean up some strings with inner quotes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,11 +80,11 @@ bdfr download ./path/to/output --user reddituser --submitted -L 100
 ```
 
 ```bash
-bdfr download ./path/to/output --user me --saved --authenticate -L 25 --file-scheme '{POSTID}'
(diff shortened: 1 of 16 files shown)


Serene-Arc:
Is bugbear a tool that we can add to our formatting list/pre-commit? Seems to be a flake8 thing but if we can add it that might be good. I haven't heard of it before.

OMEGARAZER:
It's a flake8 plugin, yes, but I've been running through everything with [Ruff](https://github.com/charliermarsh/ruff) which covers those and more and was likely going to put up a PR to integrate it within the next few days rendering its addition irrelevant. I'd just like to get reasonable coverage of its checks before adding it in.

Will open an Issue to link to the PR with the coverage that's already done to track the coverage of the checks going forward at the same time.

Serene-Arc:
Cool, I'll look at that when you do it.

PR #794, opened 2023-02-24
Partial revert of 98aa3d7

OMEGARAZER:
Reverts some quote changes for awk commands as the prints do not function the same with double quotes when the variable number is over 10

### DIFF
--- a/scripts/extract_failed_ids.sh
+++ b/scripts/extract_failed_ids.sh
@@ -8,9 +8,9 @@ else
 fi
 
 {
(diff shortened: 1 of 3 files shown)


Serene-Arc:
This does trip the tests right? We should add them to the battery that GitHub runs.

OMEGARAZER:
As bats-core doesn't run on windows they would likely need to be added separately from the rest. I briefly looked to see if there was a pre-commit for it but I didn't see one. I'll see if I can get one together.

PR #796, opened 2023-02-25
Don't create directory if not needed

OMEGARAZER:
Moves creation of parent directories after dupe check so directories are not made if not needed.


Serene-Arc:
Can you add tests that confirm this behaviour and prevent regressions?

PR #800, opened 2023-02-26
RUF lint

OMEGARAZER:
Adds coverage for RUF checks.

Adds noqa's for RUF005, as the suggested change seems to break testing on windows

Adds noqa's for RUF001 and RUF100 in formatter tests. RUF001's are expected unicode characters, RUF100 covers a Refurb noqa.


Serene-Arc:
> Adds noqa's for RUF005, as the suggested change seems to break testing on windows

Is this still an issue?

OMEGARAZER:
I haven't checked it specifically in a while but I can certainly check again.

PR #810, opened 2023-03-05
Use a more proper user agent string

OMEGARAZER:


Serene-Arc:
How does this enhance the downloader? This might cause it to stop working.

OMEGARAZER:
> This might cause it to stop working.

[It won't](https://github.com/reddit-archive/reddit/wiki/API#rules), if anything it'll probably be more likely to keep it working. I've been running with it locally for a few weeks to test with no backsteps in performance.

> Change your client's User-Agent string to something unique and descriptive, including the target platform, a unique application identifier, a version string, and your username as contact information, in the following format:
\<platform>:\<app ID>:\<version string> (by /u/\<reddit username>)
Example: User-Agent: android:com.example.myredditapp:v1.2.3 (by /u/kemitche)
Many default User-Agents (like "Python/urllib" or "Java") are drastically limited to encourage unique and descriptive user-agent strings.

This brings the user agent used for praw connections closer in line with that (except the username part obviously) since right now it just uses the hostname of the device that's running the script which may unknowingly trip up some limiting. This at least gives control over what's there and is a lot closer to what is recommended.

Serene-Arc:
Reddit isn't the only site that this will impact. We download from quite a few sites that may check the user agent. Have you tested it under high loads under a range of these?

OMEGARAZER:
These changes are only for praw connections.

What do you consider high loads? There are no test failures introduced and as I've mentioned I've been running with it for a while to make sure.

The only downloader that sets a user agent right now is Erome. Everything else just uses the base requests user agent right now. You can confirm through the JWT that [this](https://github.com/aliparlakci/bulk-downloader-for-reddit/blob/development/bdfr/site_downloaders/redgifs.py#L26-L28) gets.

just change: `assert auth1 == auth2` to `assert auth1 == ""` in the test as a simple way to get it and then you'll get something like this:
```bash
jwt-decode "eyJ0eXAiOiJKV1QiLCJhbGciOiJSUzI1NiJ9.eyJpc3MiOiJodHRwczovL3d3dy5yZWRnaWZzLmNvbS8iLCJpYXQiOjE2NzgxODM1MjAsImV4cCI6MTY3ODI2OTkyMCwic3ViIjoiY2xpZW50LzE4MjNjMzFmN2QzLTc0NWEtNjU4OS0wMDA1LWQ4ZThmZTBhNDRjMiIsInNjb3BlcyI6InJlYWQiLCJ2YWxpZF9hZGRyIjoiKiouKiouKiouKioiLCJ2YWxpZF9hZ2VudCI6InB5dGhvbi1yZXF1ZXN0cy8yLjI4LjIiLCJyYXRlIjotMX0.ABxgbHizNsgB3HRd5zatp-Hr5yWYFh-iXRxTKgSpxVv_M3e7Jubl5rQW3rZc6kek831MCEM_91jU7Ii-IkFYAWKUJCqHUxDQKxPOZTTTwd0Jx4absELYwFwmhKnF6G3SbSuE_MepiuJMEij0vD2n7CIeREOWUtFh1FJtkxj8-veAiAspks0zYOHRVJiX0HLaF4KT3mcdWh5g3GVDuZ1-ge-Df9n9k1AEr_j3LTxvkILvqB_96kGK2ZsnJ-xQ_AK-CbrWIxUV1OJKdWc31ouJyzCRYrOHCkVsLIyGbKOjhiUm7xM3DiCRgJiZhXXSiSg0jWypcK_1XpxDnT_FPkgkWA"


{
  "typ": "JWT",
  "alg": "RS256"
},
{
  "iss": "https://www.redgifs.com/",
  "iat": 1678183520,
  "exp": 1678269920,
  "sub": "client/1823c31f7d3-745a-6589-0005-d8e8fe0a44c2",
  "scopes": "read",
  "valid_addr": "**.**.**.**",
  "valid_agent": "python-requests/2.28.2",
  "rate": -1
}
```

So for this token the only user agent it will accept is `python-requests/2.28.2` or else it will say the token is invalid and throw errors which doesn't happen right now.

They certainly can be set though. Would probably be best to do something like [yt-dlp](https://github.com/yt-dlp/yt-dlp/blob/master/yt_dlp/utils.py#L75-L117) does. I can look into porting that over later though when I have a bit more time.

Serene-Arc:
Alright, I thought we used a different user agent as the default for all the site downloaders. If this is just for PRAW then it shouldn't matter, and if the other ones are working then it shouldn't be an issue at this stage.

PR #817, opened 2023-03-08
Added markdown format option

### LINKED ISSUE #816: [FEATURE] Add MarkDown as a file format option
- [x] I am requesting/proposing a feature.
- [x] I am running the latest version of BDfR
- [x] I have read the [Opening an issue](../../README.md#configuration)

## Description

I (as many people that use this awesome piece of code) am quite a data hoarder. I download _a lot_ of reddit content, mostly media. But there are also some subs that I scrape in quite regular intervals, that are text only / mainly text, e.g. r/askreddit or r/talesfromtechsupport.

But as you might expect json, yaml and/or xml are hard to read. So I had the idea that maybe an option to save posts as a simple markdown file may be an alternative to view these posts locally.

The markdown file should not contain all information of the original entry, e.g. the id or the pinned status are irrelevant in this case.

dermalikmann:
fixes #816 

This is definitly not the most beautiful code I've written, but if gets the job done. Also I've run black, and it fixed only a few lines. 

I am still not 100% sure the code for building the markdown file should be in `archiver.py` but it still felt the most appropriate place. LMK if you think it should be refactored and put somewhere else.

### DIFF
--- a/bdfr/__main__.py
+++ b/bdfr/__main__.py
@@ -65,7 +65,7 @@ _downloader_options = [
 _archiver_options = [
     click.option("--all-comments", is_flag=True, default=None),
     click.option("--comment-context", is_flag=True, default=None),
-    click.option("-f", "--format", type=click.Choice(("xml", "json", "yaml")), default=None),
+    click.option("-f", "--format", type=click.Choice(("xml", "json", "yaml", "md")), default=None),
 ]
 
 
--- a/bdfr/archiver.py
+++ b/bdfr/archiver.py
@@ -5,6 +5,7 @@ import json
 import logging
 import re
 from collections.abc import Iterable, Iterator
+from datetime import datetime
 from pathlib import Path
 from time import sleep
 from typing import Union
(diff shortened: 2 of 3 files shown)


Serene-Arc:
Thanks for your PR! The goal of the archiver function is to obtain data from Reddit about a post and store it. Both YAML and XML are markup languages designed to store data in a way that is easily understandable, even with complex data structures. This makes it easier to parse after the fact with other programs. 

Markdown doesn't have any of these features, since it is a formatting specification for simple text. I'm not sure how much use there is in a markdown option given that it doesn't have the qualities that a data language needs.

The formatting specification for each post that you've added in the code is rather arbitrary; there's no reason for it to be that way versus another, other than personal preference. I'm not sure that adding such a user-dependent option is the wisest course, in addition to there being a loss of information compared to the other methods.

If you really want to have the data in markdown text format, I would suggest that a script that parses a standard XML or YAML output would be a better way to do it than this way. We have a scripts folder that does commonly scripted tasks.

dermalikmann:
I defnitly see your point. Writing a script was actually my first instinct with this task. I just felt that such a feature would be beneficial to bdfr. I completely missed the scripts dir, sorry for that 😅 

Altho I would like to see an "integrated" way of doing stuff like this. I've skimmed over the other scripts in that dir and a quick idea that I had was, one could try to implement something like a plugin system with those. One could create a subcommand like
`python -m bdfr runscript convert-to-markdown input/dir output/dir`, which in turn would run a script with a standadized schema. What do you think of that idea? Feature Request worthy?

Anyways, im gonna close this pr and the #816 as they seem to be out of scope.

Serene-Arc:
Adding a way to implement plugins would be a wonderful addition I think. At the moment, our typical suggestion is to implement a pipeline in Bash or Powershell, which works well with log parsing and the like after runs, but being able to run scripts on the files themselves might prove useful long term.

PR #826, opened 2023-03-23
Test fixes

OMEGARAZER:


Serene-Arc:
That erome test you changed doesn't seem comparable. If the old one had 41 items and the current one has 1, that suggests a different type of page. iirc Erome allows for galleries, so the one you replaced may be that and should be replaced in kind.

OMEGARAZER:
Looking at #785 it does since it's to check to make sure it only grabs full images, not thumbnails which that link works for. With the old search that link would show 13 rather than 1.

PR #868, opened 2023-05-28
Fix Erome test

Soulsuck24:
Replaces failing erome test.


Serene-Arc:
Hey, can you add a comment describing the test case, like if it's a gallery with video or whatever, so in the future if it needs to be replaced we can replace like with like.

PR #874, opened 2023-06-06
Connector test edge case

OMEGARAZER:
Covers an edge case of not having the exact amount as the limit, and now allows up to and including the limit.

Essentially covers if posts get deleted so the account being tested drops below the limit number which is still under the limit so should be able to pass.


Serene-Arc:
I think this should have a check that fails it if there are zero submissions, if we're making this change.

PR #875, opened 2023-06-06
Ruf lint

OMEGARAZER:
#800 

RUF005 does not appear to cause issues any longer from my testing.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -89,6 +89,7 @@ markers = [
 
 [tool.ruff]
 exclude = ["scripts/tests"]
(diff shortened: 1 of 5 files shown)


Serene-Arc:
Can you add this to the pre-commit configuration? All linting and formatting tools should be in that to run before commits, and as part of the tests.

OMEGARAZER:
Ruff is in the current dev branch pre-commit

Serene-Arc:
Ah my mistake! Sorry I checked the master one by accident.

PR #886, opened 2023-06-12
Update versions

OMEGARAZER:
Updates versions of Black and Ruff.
Updates version of validate-pyproject in pre-commit.
~~Updates version of the package to 2.6.3 as it's probably overdue.~~

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -3,22 +3,23 @@
 
 repos:
   - repo: https://github.com/abravalheri/validate-pyproject
-    rev: v0.12.1
+    rev: v0.13
     hooks:
       - id: validate-pyproject
         name: validate-pyproject
 
   - repo: https://github.com/psf/black
-    rev: 23.1.0
+    rev: 23.3.0
     hooks:
       - id: black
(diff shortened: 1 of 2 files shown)


OMEGARAZER:
As I haven't gone on reddit today I forgot the blackout started today so the vast majority of the tests are going to fail on both these PR's.

Serene-Arc:
Yeah this blackout, which I support in every way, is going to be a gigantic pain regarding tests.

Please remove the bit about bumping the version. I agree that it needs to be done but that's not how versions are bumped. I make a tag, a release, it's a whole thing. I'll close a bunch of the bugs and PRs in the next week or so and then do another release, but that's a specific step that isn't right here.

OMEGARAZER:
I mean, it will need to be done before a release either way or the internal version won't be updated.

I can take it out if you want, but it will need to be done regardless.

Serene-Arc:
Please do, I would prefer to have that be a single comprehensive commit. If you haven't guessed by now, I'm somewhat intent on atomic, related commits in a PR.

OMEGARAZER:
k

PR #889, opened 2023-06-16
Oauth2 update for upcoming rate limit change

OMEGARAZER:
Change to allow the use of client id without secret.

With Reddit changing rate limiting from ID/User combinations to [solely client id](https://reddit.com/13wsiks) changing to personal client id's will likely be a required step in the future.

This PR will allow for the usage of ids without secrets that you receive when choosing "installed app" during the creation of the id.


Serene-Arc:
Did you test this to make sure it works? Due to the secrets, the online github tests won't be able to check this behaviour.

OMEGARAZER:
Yes, I've been using it with my own "installed app" id for a while, it just didn't seem to be needed here before this.

PR #898, opened 2023-06-24
Test fixes

OMEGARAZER:
Fixes tests for subreddits and posts that have either been made private or removed.

Can be reverted at a later date if need be but should fix the currently failing tests.


Serene-Arc:
Oh damn I was working on this, but you beat me to it, thanks.

PR #989, opened 2025-08-05
Download history database

rgathub:
## Description
This PR adds option to enable a sqlite database for all downloads using hash to avoid downloading the same file again. 
It also fixes the latest oAuth issue with latest version of PRAW. 

## Changes
- Fixes oAuth issue with latest version of oAuth.
- Adds flags "enable_downloads_db"  and "downloads_db_name" to create a sqlite database that stores records of downloaded files to avoid duplicate downloads in future. 
- Removed gyfcat as it does not exists from tests.
- Updated test resources as some no longer exists or have updated hashes. 

## Existing Issues
- Imgur support is out dated and is not working.


bdfr/__main__.py
```diff
@@ -60,6 +60,8 @@
     click.option("--max-score", type=int, default=None),
     click.option("--min-score-ratio", type=float, default=None),
     click.option("--max-score-ratio", type=float, default=None),
+    click.option("--enable-downloads-db", is_flag=True, default=None),
+    click.option("--downloads-db-name", type=str, default=None),
```
Serene-Arc: I think a better way to do this would not be having a flag and a DB name, but to just have something like `--downloads-db-name` where they can specify the file path. If it's `None` by default, then when providing a name, the user will also enable it.

bdfr/downloader.py
```diff
@@ -123,12 +129,9 @@ def _download_submission(self, submission: praw.models.Submission) -> None:
                 )
                 return
             resource_hash = res.hash.hexdigest()
-            if resource_hash in self.master_hash_list:
-                if self.args.no_dupes:
-                    logger.info(f"Resource hash {resource_hash} from submission {submission.id} downloaded elsewhere")
```
Serene-Arc, line 128: Logging messages should not be removed, there are tools that parse them. It can be in the refactored function, but needs to be included.

tests/site_downloaders/test_erome.py
```diff
@@ -11,27 +11,7 @@
...
-                r"https://[a-z]\d+.erome.com/\d{4}/9E50Xkb6/Ram0NmDU.jpeg",
-                r"https://[a-z]\d+.erome.com/\d{4}/9E50Xkb6/dY82guy1.jpeg",
-                r"https://[a-z]\d+.erome.com/\d{4}/9E50Xkb6/3x8bp9lF.jpeg",
-                r"https://[a-z]\d+.erome.com/\d{4}/9E50Xkb6/lxyFSUMQ.jpeg",
-                r"https://[a-z]\d+.erome.com/\d{4}/9E50Xkb6/vPIb29UR.jpeg",
-                r"https://[a-z]\d+.erome.com/\d{4}/9E50Xkb6/w1BJtyh5.jpeg",
-            ),
-        ),
-    ),
+    (("https://www.erome.com/a/vqtPuLXh", (r"https://[a-z]\d+.erome.com/\d{3}/vqtPuLXh/KH2qBT99_480p.mp4",)),),
```
Serene-Arc, line 14: If you look at the comments, these are two different types of resource. The tests must be replaced with similar ones to make sure that they don't start failing for different types of web pages. You replaced a video with a video, but there needs to be an image album as well.

bdfr/sqlite_manager.py
```diff
@@ -0,0 +1,99 @@
...
+        cursor = conn.cursor()
+        cursor.execute(
+            """
+            CREATE TABLE IF NOT EXISTS downloads (
+            id INTEGER PRIMARY KEY AUTOINCREMENT,
+            subreddit TEXT NOT NULL,
+            file_name TEXT NOT NULL,
+            file_hash TEXT NOT NULL,
+            file_size INTEGER NOT NULL,
+            timestamp DATETIME DEFAULT CURRENT_TIMESTAMP
```
Serene-Arc, line 38: Can you use the hash as the primary key? That would make more sense than using an autoincremented ID, especially since you make an index on the hash anyway.

Serene-Arc (changes requested):
Hi! Thanks for the PR. I've been meaning to go through and fix all the tests and start merging PRs again, but to be honest, I haven't been wanting to go through a bunch of porn to get all the scrapers working. Thanks for doing that for me.

This is mostly fine. I have made some notes, and if there are new options being added to the CLI, then they need to be added to the README as well.

bdfr/sqlite_manager.py
rgathub: Sure, updated the schema

bdfr/__main__.py
rgathub: agreed, consolidated the parameters.

bdfr/downloader.py
rgathub: added the message back.

PR #990, opened 2025-10-06
fix(praw): fixed auth error for new praw version

ModerateWinGuy:
Fixes this bug: https://github.com/Serene-Arc/bulk-downloader-for-reddit/issues/970


ModerateWinGuy:
Ah, I see someone's already got a PR up to fix that. I'll just close this then.

PR #994, opened 2025-12-02
Add GitHub Actions workflow for Python package publishing

szadeensfw-dot:
This workflow automates the process of uploading a Python package to PyPI when a release is created, including building the package and publishing it.


szadeensfw-dot (commented):
Need to install

PR #1012, opened 2026-04-06
Praw update

EkriirkE:
The oauth token manager is obsoleted, and removed..  This restores compatibility with the most recent praw library

### DIFF
--- a/bdfr/connector.py
+++ b/bdfr/connector.py
@@ -163,6 +163,7 @@ class RedditConnector(metaclass=ABCMeta):
                 client_id=client_id,
                 client_secret=client_secret,
                 user_agent=self.user_agent,
+                refresh_token=self.cfg_parser.get("DEFAULT", "user_token"),
                 ratelimit_seconds=120,
             )
         else:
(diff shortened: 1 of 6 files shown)


Serene-Arc:
This removes all of the code but doesn't add any. If. it's going to be fixed, we'd have to implement the new authentication, no? An unauthenticated instance would be extremely slow and throttled.

EkriirkE:
Per the notes on the last working praw BaseTokenManager https://github.com/praw-dev/praw/blob/v7.7.1/praw/util/token_manager.py the functionality was removed, and the code there just looks like an empty framework without function anyway. BDFR only fiddled with passing the token on 1 callback.  
So I see it now.  It seems praw replaced this with an explicit refresh_token vs a token_manager class with callbacks. 
Fixed as well.

--authenticate tested OK on my end with both initializing the app token, and utilizing it in subsequent calls

sirmaestro:
Without this change I couldn't get bdfr to work at all, with or without `--authenticate` flag. So I vote for adding this in.

EkriirkE:
The remaining test errors are beyond the scope of this PR (even though I did fix the iana logo download link)

PR #998, opened 2026-01-16
Fix #997, #947, #941, #774

st-korn:
Issue #997 has full description of this bug.
We can not get scopes from reddit before authorization now.

PR #1013, opened 2026-04-26
fixed praw compatibility and handling of scope endpoint errors

euocheisthai:
- the compat fix is for cases when praw 7.8+ is already installed as it happened w me; in praw 7.8 `BaseTokenManager` was moved from `praw.reddit` to `praw.util.token_manager`. Works fine w the pinned praw 7.7.1 version.
 - reddit's `/api/v1/scopes.json` endpoint now returns 403 for unauthenticated requests, which causes a `JSONDecodeError` crash before the oauth flow even starts, so a scope check is added to avoid crashing.

### DIFF
--- a/bdfr/oauth2.py
+++ b/bdfr/oauth2.py
@@ -8,6 +8,7 @@ import socket
 from pathlib import Path
 
 import praw
+import praw.util.token_manager
 import requests
 
 from bdfr.exceptions import BulkDownloaderException, RedditAuthenticationError
@@ -16,11 +17,12 @@ logger = logging.getLogger(__name__)
 
 
 class OAuth2Authenticator:
-    def __init__(self, wanted_scopes: set[str], client_id: str, client_secret: str, user_agent: str) -> None:
+    def __init__(self, wanted_scopes: set[str], client_id: str, client_secret: str, user_agent: str = "obtain_refresh_token for BDFR"):
         self._check_scopes(wanted_scopes, user_agent)
         self.scopes = wanted_scopes
         self.client_id = client_id
         self.client_secret = client_secret
+        self.user_agent = user_agent
 
     @staticmethod
     def _check_scopes(wanted_scopes: set[str], user_agent: str) -> None:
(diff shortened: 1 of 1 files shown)


vladdoster:
@euocheisthai, I just tested this out and it works for me.

khs:
Confirming the underlying bug independently — hit the same scopes.json → 403 crash on a fresh install (Windows, Python 3.13, bdfr installed from [#1012](https://github.com/Serene-Arc/bulk-downloader-for-reddit/pull/1012)'s branch since that's the praw-compat fix I found first).

But there's an issue. This PR's fix for the praw-compat half doesn't hold up on current praw. It relocates OAuth2TokenManager to subclass praw.util.token_manager.BaseTokenManager, but on praw 8.0.2 that module is gone entirely:

>>> import praw.util.token_manager
ModuleNotFoundError: No module named 'praw.util.token_manager'
So the "works for me" confirmation above was presumably on an older praw (7.8.x/7.9.x?) — the token-manager internals have already moved again since this PR was opened, which is exactly the kind of churn that keeps re-breaking this project against praw's internals.

[#1012](https://github.com/Serene-Arc/bulk-downloader-for-reddit/pull/1012) sidesteps this by dropping OAuth2TokenManager entirely and passing refresh_token= directly to praw.Reddit() — no dependency on wherever praw's internal token-manager class currently lives. Combined with this PR's _check_scopes fix (soft-failing when scopes.json 403s instead of crashing), that seems like the more durable combination. Might be worth consolidating [#1012](https://github.com/Serene-Arc/bulk-downloader-for-reddit/pull/1012) + the scopes-check fix from this PR into one PR rather than three separate ones ([#998](https://github.com/Serene-Arc/bulk-downloader-for-reddit/pull/998), [#1012](https://github.com/Serene-Arc/bulk-downloader-for-reddit/pull/1012), [#1013](https://github.com/Serene-Arc/bulk-downloader-for-reddit/pull/1013)) sitting open with overlapping scope.

